PR #32, opened 2023-08-08
Fix: DO-1576 select does not accept a derived variable with an array of strings

patricia-causalens:
## Motivation and Context

It was found that when passing `items` to `Select` component as a `DerivedVariable`, if items was not a List[Item] then the component would error. 

## Implementation Description

Added a check to see if items is a list of strings and if so to convert it into an array of items for consumption by the component. 
For NonDataVariables this conversion is done in the python side.

I have also updated prace.js to only require the first two checkboxes. I think the checkboxes should be accurate with what has been covered in the PR rather than just checking everything for the sake of it. 

None

## How Has This Been Tested?

Locally by passing a DerivedVariable with a List[Item] and one with List[string]

## PR Checklist:

### DIFF
--- a/.github/prace.yml
+++ b/.github/prace.yml
@@ -7,6 +7,6 @@ title:
     your title
 body:
   patterns:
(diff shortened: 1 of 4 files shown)


tamara-causalens (approved):

krzysztof-causalens (approved):
LGTM, could we update the docstring on the Python side to make it clear which formats are supported?

PR #34, opened 2023-08-09
Docs: DO-1629 rename best practices

tamara-causalens:
## Motivation and Context
rename best practices to design guidelines for clarity

## PR Checklist:

### DIFF
--- a/docs/sidebar.json
+++ b/docs/sidebar.json
@@ -47,7 +47,7 @@
     },
     {
       "type": "category",
-      "label": "Best Practices",
+      "label": "Design Guidelines",
       "collapsible": true,
       "collapsed": true,
       "items": [
--- a/docs/sidebar.json
+++ b/docs/sidebar.json
@@ -47,7 +47,7 @@
     },
     {
       "type": "category",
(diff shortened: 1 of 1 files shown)


patricia-causalens (approved):
What do you think about 'Tips and Techniques' instead? Sounds still suggestive and not enforcing but a bit broader so we can add things that are not design/structural specific?

docs/sidebar.json
```diff
@@ -47,7 +47,7 @@
     },
     {
       "type": "category",
-      "label": "Best Practices",
+      "label": "Design Guidelines",
```
krzysztof-causalens, line 50: I'm not sure I agree with 'Design Guidelines' as a name. Ours are more general, not necessarily only about design. Lots of frameworks use 'Best Practices' as a name, my gold standard comparison is Cypress: https://docs.cypress.io/guides/references/best-practices

tamara-causalens:
@krzysztof-causalens @patricia-causalens Would "Tips and Tricks" work better?

krzysztof-causalens (approved):

PR #35, opened 2023-08-10
Feat: DO-1503 support custom WebSocket messages

krzysztof-causalens:
## Motivation and Context

Support use cases where custom JS components need to utilise the WebSocket connection for better performance, i.e. real-time updates. Further context in Jira ticket.

## Implementation Description

This PR extends the WS system to support a new type of WS messages (both client and server side), a 'custom' message. The message format has a 'kind' attached to it, which the Python code, and the JS component can choose to handle. 

The API on the Python side looks as follows:

```python
config.add_ws_handler(kind, handler)
```

Behind the scenes the handler is added to an in-memory registry at server startup. When a custom message is encountered, the WS handling logic looks up a handler for the given kind and executes it. If the handler returns a value (other than None), a WS message is sent back with the same `kind`. 

In addition, the WS messages on the backend are now typed rather than being freeform dicts (to some extent). This is to more easily distinguish correct custom and built-in message shapes. The payloads are still mostly freeform, with the expected structures being defined in the TypeScript type definitions. This can be fully typed in the future, or perhaps some universal typing solution (JSON schema?) can be used to reduce typing duplication.

## How Has This Been Tested?

Locally in a demo app. Unit test added on the Python side.

## PR Checklist:

Example usage:

Python API

```python
class LocalJsComponent(ComponentInstance):
    content: str

config.add_component(LocalJsComponent, local=True)

def handler(chan, msg):
    print(f'Received message: {msg} from channel {chan}')
    return f'{msg}-response'

config.add_ws_handler('my_custom_kind', handler)

config.add_page(name='custom component', content=LocalJsComponent(content='test'))
```

JS component

```tsx

interface LocalJsComponentProps {
  content?: string;
}

function LocalJsComponent(props: LocalJsComponentProps) {
  const { client } = React.useContext(WebSocketCtx);
  const [state, setState] = React.useState<string>("");
  const [messages, setMessages] = React.useState<any[]>([]);

  React.useEffect(() => {
    const sub = client.customMessages$().subscribe((message) => {
      setMessages((messages) => [...messages, message]);
    });

    return () => {
      sub.unsubscribe();
    };
  }, [client]);

  return (
    <Center>
      <div>Passed Content: {props.content}</div>
      <div>Send Custom Message</div>
      <input value={state} onChange={(e) => setState(e.target.value)} />
      <button onClick={() => client.sendCustomMessage("my_custom_kind", state)}>
        Send
      </button>
      <div>Received Custom Messages</div>
      <li style={{ display: "flex", flexDirection: "column", gap: "1rem" }}>
        {messages.map((message, index) => (
          <code key={index}>{JSON.stringify(message)}</code>
        ))}
      </li>
    </Center>
  );
}
```

custom_ws

### DIFF
--- a/docs/advanced/custom-js.mdx
+++ b/docs/advanced/custom-js.mdx
@@ -500,7 +500,7 @@ from dara.core.internal.registries import utils_registry, sessions_registry, web
 # Some stored user ID
 user_id = 'some_user_id'
 
-# Get the channel associated with the user
+# Get the channel associated with the user; this lookup will only succeed if the user is currently connected
 session_id = sessions_registry.get(user_id)
 channel_id = websocket_registry.get(session_id)
 
--- a/docs/advanced/custom-js.mdx
+++ b/docs/advanced/custom-js.mdx
@@ -401,6 +401,114 @@ The action can then be used like any other built-in action.
 Button('Test button', onclick=CustomAction(country='United Kingdom', function=lambda x: f'Greetings, {x}!'))
 ```
 
+### WebSocket communication
+
+In addition to using actions or raw HTTP requests to communicate with the Dara server, custom components can utilise the WebSocket connection if real-time communication is required.
+Dara internally supports a `custom` WS message type which the application can choose to handle in any way it wants; each custom message has a `kind` field which can be used to
+distinguish between different message types.
+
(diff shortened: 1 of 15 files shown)


tamara-causalens (approved):
Looks good to me!

patricia-causalens (approved):
LGTM

PR #43, opened 2023-08-12
Fix: DO-1452 custsup table to persist variable value

patricia-causalens:
## Motivation and Context

It was difficult to persist table's selected values on page change. This is due to onclick returning the table row values and cleaning the index from it.

## Implementation Description

The proposed fix is to return the index of the columns selected as well. That way this can be extracted and used to update `selected_indices`  and keep the selection. This work similarly to other components, where the input is kept if stored in some `Variable`. 

None

## How Has This Been Tested?

By running it locally

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,28 @@
 title: Changelog
 ---
 
+## NEXT
+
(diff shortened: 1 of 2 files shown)


patricia-causalens:
Resolved in a separate PR

PR #46, opened 2023-08-14
Refactor: DO-1303 DO-1005 extended build cache - JS build process refactor

krzysztof-causalens:
## Motivation and Context

Improve the JS build process:
- made it more modular
- broken up into steps; now all information required is first gathered into the build cache file which can later be read from without having to re-import the configuration 
- removed few redundant properties
- improved static folder handling; static files are no longer copied into `static` and then into output (`dist`); instead, information on all static directories is kept all the way until the end, then all assets are copied into the output folder in order the are added
- static folders are now copied over on each server startup (fixes DO-1005)

## Implementation Description

The existing `_build.json` file was extended to hold all necessary information about the build. There is now a pydantic `BuildCache` class defining its structure. Most of the existing utilities were consolidated into instance or static methods on `BuildCache`, `JsConfig` and `BuildConfig` classes.

Added `BuildCacheDiff` class representing a set of keys in a build cache which changed across runs. Only if a key affecting the JS files has changed we then rebuild the JS, otherwise only static assets are copied.

Also fixed a few issues found in the custom JS setup:
- template config file was still `cldp.config.json` rather than `dara.config.json`
- Vite config was missing `server.fs.strict` override
- template JS file was missing React dependency

## How Has This Been Tested?

In a local project, with/without static folder(s).

## PR Checklist:


tamara-causalens (approved):
This is a very nice improvement!

PR #49, opened 2023-08-18
Docs: CAUSALAI-3268 Add license headers to python tooling scripts

andrew-causalens:
## Motivation and Context

The tooling scripts are missing the copyright and license info. Added them for the python scripts but not the TS/JS ones. I don't think this warrants a changelog entry. I also wanted to set the lower bound to `cai-causal-graph` to just be those wheels that are publicly available. This will not have an affect as the older wheels would not have been discoverable anyways. 

No

## How Has This Been Tested?

N/A

## PR Checklist:


krzysztof-causalens (approved):
LGTM, thanks!

pyproject.toml
```diff
@@ -28,7 +28,7 @@ typing-extensions = ">=4.0.0, <4.6.0"
 pyjwt = ">=2.0.1, <3.0.0"
 scipy = "*"
 fastapi-vite = "^0.3.1"
-cai-causal-graph = ">=0.0.1"
+cai-causal-graph = ">=0.1.0"
```
andrew-causalens, line 31: Wasn't sure if this was necessary or built automatically as union of each package toml.

krzysztof-causalens: Yes, this is the combined dependency range so needs to be updated if any of the packages change their deps. This can be done automatically with `borg install` as well. 

All looks good, will merge once CI passes.

PR #54, opened 2023-08-24
Feat: DO-1543 Support to lookup entry in external registry 

AidaZheng:
## Motivation and Context
Currently when render actions/derived variable/pycomponent, server will look for entry in the corresponding registry. If it doesn’t exist, will raise a KeyError.

This PR is adding a new feature: when catch a KeyError, server will check if user identified their own handler, if so, the customised handler will return the entry. 

## Implementation Description
1. Implement an `add_registry_lookup()` in the ConfigurationBuilder, user can add their own handlers.
```
from dara.core import ConfigurationBuilder

config = ConfigurationBuilder()

def get_derived_variable(uid)-> DerivedVariableRegistryEntry:
     # get external derived variable

def get_action(uid)-> Callable:
     # get external get_action

config.add_registry_lookup(
            {
                'DerivedVariable': get_derived_variable,
                'Action Handler': get_action,
            }
        )
```
2. Implement a RegistryLookup class, its take the user defined handlers when initialise and has a get function, which takes a registry and a uid as params. When the uid is not in the registry, it will call the lookup handler customised by user. 
```
from dara.core.internal.registry_lookup import RegistryLookup
from dara.core.internal.registries import derived_variable_registry

registry_mgr: RegistryLookup = utils_registry.get('RegistryLookup')
variable = await registry_mgr.get(derived_variable_registry, uid)
```
None

## How Has This Been Tested?
Manually tested

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/configuration.py
+++ b/packages/dara-core/dara/core/configuration.py
@@ -42,6 +42,7 @@ from dara.core.internal.import_discovery import (
     create_component_definition,
     run_discovery,
 )
+from dara.core.internal.registry_lookup import CustomRegistryLookup
 from dara.core.internal.scheduler import ScheduledJob, ScheduledJobFactory
 from dara.core.logging import dev_logger
 from dara.core.visual.components import RawString
@@ -52,7 +53,7 @@ class Configuration(GenericModel):
     """Definition of the main framework configuration"""
 
     auth_config: BaseAuthConfig
-    registry_lookup: Dict[str, Callable]
+    registry_lookup: CustomRegistryLookup
     actions: List[ActionDef]
     endpoint_configurations: List[EndpointConfiguration]
     components: List[ComponentTypeAnnotation]
--- a/packages/dara-core/dara/core/configuration.py
+++ b/packages/dara-core/dara/core/configuration.py
@@ -110,7 +110,7 @@ class ConfigurationBuilder:
     """
 
     auth_config: BaseAuthConfig
-    registry_lookup: Dict[str, Callable]
+    registry_lookup: CustomRegistryLookup
     _actions: List[ActionDef]
     _components: List[ComponentTypeAnnotation]
     _errors: List[str]
@@ -371,7 +371,7 @@ class ConfigurationBuilder:
 
         return auth
 
-    def add_registry_lookup(self, registry_lookup: Dict[str, Callable[[str], Coroutine]]):
+    def add_registry_lookup(self, registry_lookup: CustomRegistryLookup):
         """
         Register custom external registry lookup handlers, which will be called when a uid its not in the server registry
         Example:
--- a/packages/dara-core/dara/core/configuration.py
+++ b/packages/dara-core/dara/core/configuration.py
@@ -18,7 +18,7 @@ limitations under the License.
 import os
 import pathlib
 from types import ModuleType
-from typing import Any, Callable, Dict, List, Literal, Optional, Set, Tuple, Type, Union
+from typing import Any, Callable, Coroutine, Dict, List, Literal, Optional, Set, Tuple, Type, Union
 
 from pydantic.generics import GenericModel
 
@@ -370,7 +370,28 @@ class ConfigurationBuilder:
 
         return auth
 
-    def add_registry_lookup(self, registry_lookup: Dict):
+    def add_registry_lookup(self, registry_lookup: Dict[str, Callable[[str], Coroutine]]):
+        """
+        Register custom external registry lookup handlers, which will be called when a uid its not in the server registry
+        Example:
+
+        ```python
+        from dara.core import ConfigurationBuilder
+        config = ConfigurationBuilder()
+
+        def get_derived_variable(uid)-> DerivedVariableRegistryEntry:
+            # return derived variable
+
+        def get_action(uid)-> Callable:
+            # return action
+
+        config.add_registry_lookup(
+                    {
+                        'DerivedVariable': get_derived_variable,
+                        'Action Handler': get_action,
+                    }
+                )
+        """
         self.registry_lookup = registry_lookup
 
         return registry_lookup
--- a/packages/dara-core/dara/core/configuration.py
+++ b/packages/dara-core/dara/core/configuration.py
@@ -18,7 +18,7 @@ limitations under the License.
 import os
 import pathlib
 from types import ModuleType
-from typing import Any, Callable, Coroutine, Dict, List, Literal, Optional, Set, Tuple, Type, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Set, Tuple, Type, Union
 
 from pydantic.generics import GenericModel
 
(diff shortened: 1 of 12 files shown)


packages/dara-core/dara/core/configuration.py
```diff
@@ -367,6 +370,11 @@ def add_auth(self, auth: BaseAuthConfig):
 
         return auth
 
+    def add_registry_lookup(self, registry_lookup: Dict):
```
krzysztof-causalens, line 373: Could we add a docstring?

packages/dara-core/js/shared/utils/use-component-registry.tsx
```diff
@@ -14,27 +16,31 @@ interface ComponentRegistryInterface {
...
+                        const { data } = await refetchComponents();
+                        registry = data;
+                        if (registry && registry[instance.name]) {
+                            return registry[instance.name];
+                    }
+                    i++;
                 }
-
-                i++;
             }
```
krzysztof-causalens: I think the logic here could be simplified a little bit, now you have duplication where you have this bit

```
if (registry && registry[instance.name]) {
   return registry[instance.name];
```

twice, once before the loop and once within. Could we contain all logic within the loop, and let's add some comments. The rough flow here should be:
- For each iteration (until max retries)
1) check if value is already in registry, return if that's the case and quit the loop
2) otherwise make a direct request for the component (invoking custom lookup potentially)
3) refetch the registry regardless of whether 2) succeeded, just to keep our 
  local registry copy fresh to prevent future requests
4) if we found a component in 2), return it

packages/dara-core/dara/core/internal/registry_lookup.py
```diff
@@ -0,0 +1,42 @@
...
+        self.handlers = handlers
+
+    async def get(self, registry: Registry, uid: str):
+        try:
+            return registry.get(uid)
+        except KeyError as e:
+            if registry.name in self.handlers:
+                func = self.handlers[registry.name]
+                entry = await func(uid)
+                registry.register(uid, entry)
```
krzysztof-causalens, line 38: You might want to first remove the old entry in case the registry does not permit overriding entries (Registry class has a 'allow_duplicates' flag which is False for some registries).

packages/dara-core/dara/core/internal/registry_lookup.py
```diff
@@ -0,0 +1,42 @@
...
+
+class RegistryLookup:
+    """
+    Manages registry Lookup.
+    """
+
+    def __init__(self, handlers: Optional[Dict] = {}):
+        self.handlers = handlers
+
+    async def get(self, registry: Registry, uid: str):
```
krzysztof-causalens, line 31: Let's add a docstring here

packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
```diff
@@ -321,6 +321,13 @@ function PythonWrapper(props: PythonWrapperProps): JSX.Element {
     }
 
     if (isRawString(component)) {
+        let res = component.props.content
+        if(res.startsWith('__dara__')){
+            res = res.slice(8);
+            const resJson = JSON.parse(res);
+            console.log(resJson);
+            return <DynamicComponent component={resJson} key={component.uid} />;
+        }
```
krzysztof-causalens: I'm not sure this logic belongs here, can this be handled on the app side? Happy to discuss privately

packages/dara-core/dara/core/internal/dependency_resolution.py
```diff
@@ -97,9 +98,11 @@ async def _resolve_derived_data_var(entry: ResolvedDerivedDataVariable, store: S
     from dara.core.internal.registries import (
         data_variable_registry,
         derived_variable_registry,
+        utils_registry,
     )
 
-    dv_var = derived_variable_registry.get(str(entry.get('uid')))
+    registry_mgr: RegistryLookup = utils_registry.get('RegistryLookup')
+    dv_var = await registry_mgr.get(derived_variable_registry, str(entry.get('uid')))
     data_var = data_variable_registry.get(str(entry.get('uid')))
```
krzysztof-causalens, line 106: We should also handle the data variable registry everywhere

packages/dara-core/dara/core/configuration.py
```diff
@@ -52,6 +52,7 @@ class Configuration(GenericModel):
     """Definition of the main framework configuration"""
 
     auth_config: BaseAuthConfig
+    registry_lookup: Dict[str, Callable]
```
krzysztof-causalens, line 55: If we're going with the dict approach, could we define an enum for the registry names which we currently support for the custom registry lookup?

e.g. in the registries file have an enum

```python
class RegistryType(str, Enum):
   FOO = 'Foo'
   BAR = 'Bar'
   BAZ = 'Baz'
```

Then in registry lookup we can define which registries we support custom lookup for; otherwise it's not clear which ones work

```python
RegistryLookupKey = Literal[RegistryType.FOO, RegistryType.BAR]
CustomRegistryLookup = Dict[RegistryLookupKey, Callable[<some coroutine type>]]
```

This way we can specify which handlers are supported for the registry lookup

```python
def add_registry_lookup(self, registry_lookup: CustomRegistryLookup):
```

krzysztof-causalens, line 55: Also let's update the type everywhere to specify the Callable must return a coroutine, if when we call it in RegistryLookup we always `await` it. This way users know they have to provide an `async` function

krzysztof-causalens (changes requested):
Looks good in general, well done! Just a few comments

packages/dara-core/dara/core/internal/routing.py
```diff
@@ -170,7 +167,17 @@ async def get_auth_components():  # pylint: disable=unused-variable
...
+    async def get_components(name: Optional[str]=None):  # pylint: disable=unused-variable
+        """
+        If name is passed, will only return the specific component
+
+        :param name: the name of component
+        """
+        if name is not None:
+            registry_mgr: RegistryLookup = utils_registry.get('RegistryLookup')
+            comp = await registry_mgr.get(component_registry, name)
+            return {'name': comp.dict(exclude={'func'})}
```
krzysztof-causalens, line 179: Actually if you returned the entire registry here as well, you could skip the second refetch on the frontend. You're not using the returned single component anyway? 
Though make sure to handle the component being missing in the registry, we probably want to return the entire registry anyway no matter what.

```tsx
// first time try asking directly for the missing component
if (i === 0) {
    
    const res = await request(`/api/core/components?name=${instance.name}`, { method: HTTP_METHOD.GET }, token);
    registry = await res.json();
} else {
  // otherwise just keep retrying the entire registry - existing code
  
}
i++;
```

krzysztof-causalens (approved):
Looks good, nice job! Let's just run the formatters (`borg run format`) and make sure the CI passes, and good to merge!

tamara-causalens (approved):
Lovely, feel free to merge if it is good to go!

PR #56, opened 2023-08-29
Improvement: DO-1680 internal auth logic consistency

krzysztof-causalens:
## Motivation and Context

As part of debugging auth issues in embedded environments (i.e. inside an iframe) a few inconsistencies in internal auth logic were spotted.
Some places were still using the `DARA_JWT_TOKEN` const instead of checking whether they should use the pageId key.

## Implementation Description

The logic within `dara.core` has been cleaned up, all places using the token are now going through a centralized `getToken` or `getTokenKey` method which check which key to look up the token at. 

Also fixed too narrow typing on `useVariable`.

Including build for 1.1.2.

## How Has This Been Tested?

Tested the logic still works for basic/implicit auth in non-embedded env and stores under the correct global key, embedded environment stores under provided page id. Full embedded testing TBD pending further debugging of aforementioned issue (see Jira ticket).

Standard app
Screenshot 2023-08-29 at 14 58 53
Embedded
Screenshot 2023-08-29 at 15 03 23

## PR Checklist:

### DIFF
--- a/borg.toml
+++ b/borg.toml
@@ -1,1 +1,1 @@
-version = "1.1.1"
+version = "1.1.2"
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "npmClient": "pnpm",
-  "version": "1.1.1",
+  "version": "1.1.2",
   "packages": [
     "packages/*"
   ],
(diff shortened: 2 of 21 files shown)


patricia-causalens (approved):
LGTM

PR #65, opened 2023-09-11
Fix: DO-1755: Fix static assets path propagation

Roman-Kornev:
## Motivation and Context
The logo path was not resolving correctly.

## Implementation Description
`dara.window.base_url` is now appended to the asset URL in all cases.

## How Has This Been Tested?
Locally.

## PR Checklist:

BEFORE:
image
AFTER:
image

### DIFF
--- a/packages/dara-core/js/components/side-bar-frame/side-bar-frame.tsx
+++ b/packages/dara-core/js/components/side-bar-frame/side-bar-frame.tsx
@@ -1,13 +1,14 @@
+import { transparentize } from 'polished';
+
+import styled, { ThemeContext, useTheme } from '@darajs/styled-components';
+import { Button } from '@darajs/ui-components';
+
 import { useConfig } from '@/api';
 import DaraDark from '@/assets/dara-dark.svg';
 import DaraLight from '@/assets/dara-light.svg';
 import { DirectionCtx, DynamicComponent, Wrapper, getIcon, resolveTheme } from '@/shared';
 import { ComponentInstance } from '@/types';
 import { prependBaseUrl } from '@/utils';
-import { transparentize } from 'polished';
-
-import styled, { ThemeContext, useTheme } from '@darajs/styled-components';
-import { Button } from '@darajs/ui-components';
 
 interface SideBarProps {
     width?: string;
--- a/packages/dara-core/js/components/side-bar-frame/side-bar-frame.tsx
+++ b/packages/dara-core/js/components/side-bar-frame/side-bar-frame.tsx
@@ -8,6 +8,7 @@ import DaraDark from '@/assets/dara-dark.svg';
 import DaraLight from '@/assets/dara-light.svg';
 import { DirectionCtx, DynamicComponent, Wrapper, getIcon, resolveTheme } from '@/shared';
 import { ComponentInstance } from '@/types';
+import { prependBaseUrl } from '@/utils';
 
 interface SideBarProps {
     width?: string;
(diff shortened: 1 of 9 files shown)


krzysztof-causalens (changes requested):
That's not quite how the image path resolving works AFAIK.
The reason it's `/static`, is because it refers to the URL of the image as served by the FastAPI server; `/static` is where the output folder (i.e. usually `dist`) is served under. Specifying `./static` shouldn't work.

Was there an issue which sparked this PR? This might be a red herring and the issue could be elsewhere. Happy to discuss in detail outside the PR

packages/dara-core/js/components/side-bar-frame/side-bar-frame.tsx
```diff
@@ -107,9 +108,9 @@ const LogoutArrow = getIcon('fa-solid fa-arrow-right-from-bracket');
 function SideBarFrame(props: SideBarFrameProps): JSX.Element {
     const theme = useTheme();
     const { data: config } = useConfig();
-    const logo = props.logo_path && <LogoImage alt="Logo" src={props.logo_path} width={props.logo_width} />;
-    const daraLogo =
-        theme.themeType === 'dark' ? <img alt="Dara" src={DaraDark} /> : <img alt="Dara" src={DaraLight} />;
+    const logo = props.logo_path && <LogoImage alt="Logo" src={prependBaseUrl(props.logo_path)} width={props.logo_width} />;
+    const logoSrc = theme.themeType === 'dark' ? DaraDark : DaraLight;
+    const daraLogo = <img alt="Dara" src={prependBaseUrl(logoSrc)} />
```
krzysztof-causalens, line 113: There's two logos here:
- the Dara logo (Dark or Light) does not need prepending as it's imported directly as an asset
- the `props.logo_path` logo is the one that does need prepending

So you can remove the prepend from logoSrc

packages/dara-core/js/components/top-bar-frame/top-bar-frame.tsx
```diff
@@ -114,9 +114,11 @@ const LogoutArrow = getIcon('fa-solid fa-arrow-right-from-bracket');
...
     const theme = useTheme();
     const { data: config } = useConfig();
-    const logo = props.logo_path && <LogoImage alt="Logo" src={props.logo_path} width={props.logo_width} />;
-    const daraLogo =
-        theme.themeType === 'dark' ? <img alt="Dara" src={DaraDark} /> : <img alt="Dara" src={DaraLight} />;
+    const logo = props.logo_path && (
+        <LogoImage alt="Logo" src={prependBaseUrl(props.logo_path)} width={props.logo_width} />
+    );
+    const logoSrc = theme.themeType === 'dark' ? DaraDark : DaraLight;
+    const daraLogo = <img alt="Dara" src={prependBaseUrl(logoSrc)} />;
```
krzysztof-causalens, line 121: Same here

packages/dara-core/js/utils.ts
```diff
@@ -0,0 +1,6 @@
+export const prependBaseUrl = (asset: string): string => {
+    if (window.dara?.base_url) {
+        return new URL(asset, window.dara.base_url).toString();
```
krzysztof-causalens, line 3: Just checking, did you verify this works in our use cases?

Roman-Kornev:
Turns out the issue was that we were not prepending the base_url in some cases. Reverted the changes and pushed a different fix.

packages/dara-core/js/components/side-bar-frame/side-bar-frame.tsx
Roman-Kornev: Removed prepending from `daraLogo` in both cases.

krzysztof-causalens (approved):

PR #74, opened 2023-09-18
Fix: DO-1452, DO-1713 fix table sort does not return the right index row

patricia-causalens:
## Motivation and Context

A bug was found where if table was sorted then the wrong chosen role was passed to the backend.

## Implementation Description

The issue turned out to be an issue with getting the value within the getItem which created a discrepancy when data was sorted from initial index to the one chosen by the user. This was also an issue for if the table selected value was changed from outside the component. The easier fix was to simply remove that part of the logic. We do need to fetch the data more often, but in this way we guarantee that regardless if selection is done in a sorted table or changed from outside that the correct index is still passed to both `onclick_row` action context and `selected_indices` Variables.

None

## How Has This Been Tested?

Locally by single and multi selecting table rows, sorted and unsorted and checking corresponding Variables were updated as expected.

## PR Checklist:

### DIFF
--- a/packages/dara-components/js/common/table/table.tsx
+++ b/packages/dara-components/js/common/table/table.tsx
@@ -373,12 +373,14 @@ function Table(props: TableProps): JSX.Element {
 
     const datetimeColumns = useMemo(() => getDatetimeColumns(resolvedColumns), [resolvedColumns]);
     const fetchData = useCallback(
-        async (startIndex: number, stopIndex: number) => {
+        async (startIndex: number, stopIndex: number, index?: number) => {
             const response = await getData(combineFilters('AND', [filtersToFilterQuery(filters), searchQuery]), {
+                index,
                 limit: stopIndex - startIndex,
-                offset: startIndex,
+                offset: sortingRules[0] ? startIndex + 5 : startIndex,
                 sort: sortingRules[0],
             });
+
             return {
                 data: response.data.map((row: DataRow) => {
                     for (const val of datetimeColumns) {
@@ -408,7 +410,7 @@ function Table(props: TableProps): JSX.Element {
     async function getRowByIndex(idx: number): Promise<DataRow> {
         // populate cache with a few rows around the index if row not in cache
         if (!extraDataCache.current[idx]) {
-            const { data } = await fetchData(idx - 5, idx + 5);
+            const { data } = await fetchData(idx - 5, idx + 5, idx);
             for (const row of data) {
                 extraDataCache.current[row[INDEX_COL]] = row;
             }
--- a/packages/dara-components/js/common/table/table.tsx
+++ b/packages/dara-components/js/common/table/table.tsx
@@ -410,7 +410,7 @@ function Table(props: TableProps): JSX.Element {
     async function getRowByIndex(idx: number): Promise<DataRow> {
         // populate cache with a few rows around the index if row not in cache
         if (!extraDataCache.current[idx]) {
-            const { data } = await fetchData(idx);
+            const { data } = await fetchData(undefined, undefined, idx);
             for (const row of data) {
                 extraDataCache.current[row[INDEX_COL]] = row;
             }
--- a/packages/dara-components/js/common/table/table.tsx
+++ b/packages/dara-components/js/common/table/table.tsx
@@ -373,9 +373,11 @@ function Table(props: TableProps): JSX.Element {
 
     const datetimeColumns = useMemo(() => getDatetimeColumns(resolvedColumns), [resolvedColumns]);
     const fetchData = useCallback(
-        async (index?: number) => {
+        async (startIndex?: number, stopIndex?: number, index?: number) => {
             const response = await getData(combineFilters('AND', [filtersToFilterQuery(filters), searchQuery]), {
                 index,
+                limit: stopIndex !== undefined && startIndex !== undefined ? stopIndex - startIndex : undefined,
+                offset: startIndex !== undefined ? startIndex : undefined,
                 sort: sortingRules[0],
             });
 
--- a/packages/dara-components/js/common/table/table.tsx
+++ b/packages/dara-components/js/common/table/table.tsx
@@ -401,23 +401,11 @@ function Table(props: TableProps): JSX.Element {
 
     /**
      * Returns a row by index
-     * - first tries to retrieve it from the loader data
-     * - if not found, tries to retrieve it from the extraDataCache
-     * - if not found, fetches a few rows around the index and stores them in the extraDataCache
+     * Fetches a few rows around the index and stores them in the extraDataCache
      *
      * @param idx index of row to get
      */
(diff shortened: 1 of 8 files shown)


patricia-causalens:
@krzysztof-causalens looked at the error in the CI and their docs recommend to sanitize the data beforehand, is there a reason why we aren't?

krzysztof-causalens:
> @krzysztof-causalens looked at the error in the CI and their docs recommend to sanitize the data beforehand, is there a reason why we aren't?

Sanitizing is mostly important where the HTML comes from the user. Yes ideally we'd sanitize it but "fine" since it's provided by the app developer (to the best of my knowledge, not a security expert), hence the error was ignored. Not sure why suddenly the bearer-ignore stopped working, perhaps they changed how their ignore system works and the comment needs updating?

packages/dara-components/js/common/table/table.tsx
```diff
@@ -401,23 +401,11 @@ function Table(props: TableProps): JSX.Element {
...
     /**
      * Returns a row by index
-     * - first tries to retrieve it from the loader data
-     * - if not found, tries to retrieve it from the extraDataCache
-     * - if not found, fetches a few rows around the index and stores them in the extraDataCache
+     * Fetches a few rows around the index and stores them in the extraDataCache
      *
      * @param idx index of row to get
      */
     async function getRowByIndex(idx: number): Promise<DataRow> {
```
krzysztof-causalens, line 408: Interesting, we previously discussed a solution which would explicitly fetch rows every time, utilizing the existing cache is definitely more efficient. How does it work with sorting and filtering? Does it always yield the right rows because of the cache being cleared when `getData` changes?

packages/dara-components/js/common/table/table.tsx
```diff
@@ -373,12 +373,14 @@ function Table(props: TableProps): JSX.Element {
 
     const datetimeColumns = useMemo(() => getDatetimeColumns(resolvedColumns), [resolvedColumns]);
     const fetchData = useCallback(
-        async (startIndex: number, stopIndex: number) => {
+        async (startIndex: number, stopIndex: number, index?: number) => {
             const response = await getData(combineFilters('AND', [filtersToFilterQuery(filters), searchQuery]), {
+                index,
                 limit: stopIndex - startIndex,
-                offset: startIndex,
+                offset: sortingRules[0] ? startIndex + 5 : startIndex,
```
krzysztof-causalens, line 380: Could you explain what this does and why we're modifying the offset at all?

packages/dara-core/dara/core/interactivity/filtering.py
```diff
@@ -38,6 +38,7 @@ class Pagination(BaseModel):
     offset: Optional[int] = None
     limit: Optional[int] = None
     orderBy: Optional[str] = None
+    index: Optional[str] = None
```
krzysztof-causalens, line 41: Let's update the docstring to explain how `index` affects the returned results

packages/dara-components/js/common/table/table.tsx
patricia-causalens: That was the idea, ended up not fully working on further testing, so added the index as we had previously discussed

patricia-causalens:
> > @krzysztof-causalens looked at the error in the CI and their docs recommend to sanitize the data beforehand, is there a reason why we aren't?
> 
> Sanitizing is mostly important where the HTML comes from the user. Yes ideally we'd sanitize it but "fine" since it's provided by the app developer (to the best of my knowledge, not a security expert), hence the error was ignored. Not sure why suddenly the bearer-ignore stopped working, perhaps they changed how their ignore system works and the comment needs updating?

Note for future: The message was relating to a Bearer issue that is currently up, meantime workaround was found in a separate PR and added here.

packages/dara-components/js/common/table/table.tsx
patricia-causalens: Oops not needed, I was just testing and forgot to return to normal before the commit, will fix

packages/dara-core/dara/core/interactivity/filtering.py
patricia-causalens: Good point, updated

patricia-causalens:
Checked and with persist_value flag the table selection persists, tested for an initial value selected and for a combination of filter + sort

packages/dara-components/js/common/table/table.tsx
```diff
@@ -401,26 +403,14 @@ function Table(props: TableProps): JSX.Element {
...
-            return loaderItem;
-        }
-
-        // This should only happen if user changes selectedRow from the outside
-        // and the row has not been loaded yet
-
         // populate cache with a few rows around the index if row not in cache
         if (!extraDataCache.current[idx]) {
-            const { data } = await fetchData(idx - 5, idx + 5);
+            const { data } = await fetchData(idx - 5, idx + 5, idx);
```
krzysztof-causalens, line 413: Do we even need to pass the indices (-5, +5) here? Since we pass the exact index we want to fetch anyway

krzysztof-causalens (approved):
Looks good, nice job! final minor suggestion below

tamara-causalens (approved):

packages/dara-components/js/common/table/table.tsx
patricia-causalens: removed

packages/dara-components/js/common/table/table.tsx
```diff
@@ -373,12 +373,12 @@ function Table(props: TableProps): JSX.Element {
 
     const datetimeColumns = useMemo(() => getDatetimeColumns(resolvedColumns), [resolvedColumns]);
     const fetchData = useCallback(
-        async (startIndex: number, stopIndex: number) => {
+        async (index?: number) => {
```
krzysztof-causalens, line 376: I think this function still needs to accept the start/end index for the infinite loader to work correctly. I was just referring to the `getRowByIndex` not having to pass in those indices.

krzysztof-causalens (changes requested):

packages/dara-components/js/common/table/table.tsx
patricia-causalens: Oh I see, misunderstood what you meant, fixed it now.

packages/dara-components/js/common/table/table.tsx
```diff
@@ -401,26 +403,14 @@ function Table(props: TableProps): JSX.Element {
...
-            return loaderItem;
-        }
-
-        // This should only happen if user changes selectedRow from the outside
-        // and the row has not been loaded yet
-
         // populate cache with a few rows around the index if row not in cache
         if (!extraDataCache.current[idx]) {
-            const { data } = await fetchData(idx - 5, idx + 5);
+            const { data } = await fetchData(idx);
```
krzysztof-causalens, line 413: This will now pass in `idx` as `startIndex`, not as `index`

krzysztof-causalens (approved):

PR #79, opened 2023-09-21
Feat: DO-1785 Add serialization support

AidaZheng:
## Motivation and Context
Errors occur when the backend is trying to return a component json data to the frontend, which contains datatypes in pandas/numpy. The reason behind this is fastapi doesn't know how to handle serialisation of those datatypes.

Hence, an encoder registry is introduced to handle the serialisation and deserialisation.

## Implementation Description
Defined a global encoder dict registry, in which key is a class Type, and each type have two handlers, one for serialisation and one for serialisation.

User can add their custom encoder to the global encoder dict registry by calling config.add_encoder()

None

## How Has This Been Tested?
Locally tested.

## PR Checklist:

### DIFF
--- a/docs/best-practices/retaining-correct-variable-types.md
+++ b/docs/best-practices/retaining-correct-variable-types.md
@@ -1,6 +1,40 @@
 ---
 title: Retaining Correct Variable Types
 ---
+### Adding Type Annotations to Your Resolver
+`dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. When a `Variable` is passed to one of these, it is serialized into a format that FastApi can handle and becomes a python basic type format on the server side. Eg, a  `numpy.array` will became a  `list`, a `numpy.int8` will become a `int`.
+
+
+By default, Dara supports to serialize all the generic data types in Numpy. See [default data types supported](https://github.com/causalens/dara/blob/master/packages/dara-core/dara/core/internal/encoder_registry.py).
+
+You can also add your custom encoder by using `ConfigurationBuilder.add_encoder()`. Notice, if a Variable is a type that can not be serialized by either default encoder handler or custom encoder handle, it will case a crush.
+```python
+from dara.core import ConfigurationBuilder
+
+config = ConfigurationBuilder()
+config.add_encoder(typ=np.array, serialize=lambda x: x.tolist(), deserialize=lambda y: np.array(y))
+```
+
+As mentioned above, the Variable will be serialized into a format that FastApi can handle and you can deserialize it by adding type annotations to your function.
+```python
+from dara.core import DerivedVariable, Variable,py_component
+import numpy
+
+my_var = Variable(default=numpy.array([1,2,3]))
+
+def dv_func(var: numpy.array):
+  # The var is a numpy.array, you can use the numpy array function to manipulate data
+  return var.sum()
+
+my_der_var = DerivedVariable(dv_func, variables=[my_var])
+
+# Same for py_component
+@py_component
+def my_component(var: numpy.array):
+    return Text(var.tobytes())
+my_component(my_var)
+```
+Currently the serialize/deserialize handler does not work for action resolvers due to their different API shape, which we'll hopefully solve in the near future.
 
 ### Using pydantic
 `dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. When a `Variable` is passed to one of these, it is serialized into a JSON format which becomes a plain `dict` on the server side.
@@ -45,7 +79,7 @@ from externalpackage import ExternalClass
 class MyClass(ExternalClass):
   def from_dict(self, dict):
       ...
-      
+
 my_var = Variable(MyClass(foo='a'))
 
 # A dict will be passed to resolving function
@@ -53,6 +87,6 @@ def bar(some_var: dict):
   # use from_dict to get back MyClass instance
   var_value = MyClass.from_dict(some_var)
   return var_value.foo
-  
+
 my_der_var = DerivedVariable(bar, variables=[my_var])
 ```
--- a/docs/best-practices/retaining-correct-variable-types.md
+++ b/docs/best-practices/retaining-correct-variable-types.md
@@ -2,12 +2,17 @@
 title: Retaining Correct Variable Types
 ---
 ### Adding Type Annotations to Your Resolver
-`dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. When a `Variable` is passed to one of these, it is serialized into a format that FastApi can handle and becomes a python basic type format on the server side. Eg, a  `numpy.array` will became a  `list`, a `numpy.int8` will become a `int`.
+`dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. This means that there are times when Dara needs to be able to serialize and deserialize your values.
 
+Static values or `Variable` default values need to be serialized to JSON as part of the component tree - either when included directly in the page or when returned from a `py_component`.
 
-By default, Dara supports to serialize all the generic data types in Numpy. See [default data types supported](https://github.com/causalens/dara/blob/master/packages/dara-core/dara/core/internal/encoder_registry.py).
+When a user-defined resolver (for `DerivedVariable`s, `py_component`s or action) is called, Dara receives the `Variable` values from the frontend as the serialized JSON. This means your data structures would be turned into primitive types, a `numpy.array` will became a `list`, a `numpy.int8` will become a `int`.
 
-You can also add your custom encoder by using `ConfigurationBuilder.add_encoder()`. Notice, if a Variable is a type that can not be serialized by either default encoder handler or custom encoder handle, it will case a crush.
+To preserve your original types, Dara provides an encoder system which will attempt to turn the primitive types back into the original type, as long as you annotate your resolver with a type for which an encoder is defined.
+
+Out of the box, Dara comes with encoders for all generic data types in `pandas` and `numpy`. . See [default data types supported](https://github.com/causalens/dara/blob/master/packages/dara-core/dara/core/internal/encoder_registry.py).
+
+You can also add your custom encoder by using `ConfigurationBuilder.add_encoder()`. Notice, if a Variable is a type that can not be serialized by either default encoder handler or custom encoder handle, it can cause serialization to fail.
 ```python
 from dara.core import ConfigurationBuilder
 
@@ -34,7 +39,7 @@ def my_component(var: numpy.array):
     return Text(var.tobytes())
 my_component(my_var)
 ```
-Currently the serialize/deserialize handler does not work for action resolvers due to their different API shape, which we'll hopefully solve in the near future.
+Current limitation of the serialize/deserialize handler is that the automatic deserialization does not work for action resolvers due to their different API shape.
 
 ### Using pydantic
 `dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. When a `Variable` is passed to one of these, it is serialized into a JSON format which becomes a plain `dict` on the server side.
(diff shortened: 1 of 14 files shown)


tamara-causalens:
Hey @AidaZheng can we also add documentation for this feature please (you can add it in advanced section) as well as a note in the changelog? Thank you!

packages/dara-core/dara/core/interactivity/derived_variable.py
```diff
@@ -207,7 +208,9 @@ def _restore_pydantic_models(func: Callable[..., Any], *args):
                 if typ is not None and isclass(typ) and issubclass(typ, BaseModel) and arg is not None:
                     parsed_args.append(typ(**arg))
                     continue
```
krzysztof-causalens, line 210: I think user-defined encoders should take precedence over our catch-all BaseModel handler. In some cases a BaseModel subclass might have a custom initializer where the default approach of `typ(**arg)` might not work.

Same in the other place restoring arg annotations

packages/dara-core/dara/core/main.py
```diff
@@ -69,6 +71,10 @@
 )
 from dara.core.logging import LoggingMiddleware, dev_logger, eng_logger, http_logger
 
+# Inject the encoder to pydantic ENCODERS_BY_TYPE, which will be called in fastapi jsonable_encoder
+for key, value in encoder_registry.items():
+    ENCODERS_BY_TYPE[key] = value['serialize']
+
```
krzysztof-causalens, line 77: Shouldn't this be called within the startup function so it is filled in after we fill our `encoder_registry` with user defined encoders?

packages/dara-core/dara/core/interactivity/derived_variable.py
```diff
@@ -207,7 +208,9 @@ def _restore_pydantic_models(func: Callable[..., Any], *args):
...
-                parsed_args.append(arg)
+                elif typ in encoder_registry:
+                    parsed_args.append(encoder_registry[typ]['deserialize'](arg))
+                    continue
```
krzysztof-causalens, line 213: I think this loop is missing the `parsed_args.append(arg)` statement, if neither condition is satisfied we need to return the arg as-is

packages/dara-core/dara/core/visual/dynamic_component.py
```diff
@@ -255,6 +256,8 @@ async def render_component(
             typ = annotations.get(key)
             if typ is not None and isclass(typ) and issubclass(typ, BaseModel) and isinstance(value, dict):
                 val = typ(**value)
+            elif typ in encoder_registry:
+                val = encoder_registry[typ]['deserialize'](val)
```
krzysztof-causalens, line 260: Same here, reverse condition order

krzysztof-causalens (commented):
Looks good, nice job. Spotted a few improvements below.

Agreed with Tamara about docs. We currently have a page in `dara/docs/best-practices/retaining-correct-variable-types.md` which talks about restoring pydantic types. Let's adopt it and:
- explain the serialize/deserialize problem, 
- how to use this API to solve it, and how type annotations on resolvers are required
- the built-in encoders we provide,
- the fallback BaseModel handling we do

Maybe you can also note this currently does not work for action resolvers due to their different API shape (which we'll hopefully solve in the near future).

packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Construct numpy generic datatype
+
+    :param typ: The numpy generic datatype class
+    """
+    return Encoder(serialize=lambda x: x.item(), deserialize=lambda x: typ(x))
+
+
+# A encoder_registry to handle serialization/deserialization for numpy/pandas type
+encoder_registry: MutableMapping[Type[Any], Encoder] = {
```
krzysztof-causalens, line 39: @tamara-causalens can you check whether this covers most use common data science use cases (for the open source). Internal type serialization can be added to our internal plugin

AidaZheng: > @tamara-causalens can you check whether this covers most use common data science use cases (for the open source). Internal type serialization can be added to our internal plugin

I handle all the subclass of generic type in Numpy. In pandas DataFrame, the datatype could a numpy dtype or pyarrow dtype. Let me know if we should handle the pyarrow as well.
```
<class 'numpy.number'>
     <class 'numpy.integer'>
         <class 'numpy.signedinteger'>
             <class 'numpy.int8'>
             <class 'numpy.int16'>
             <class 'numpy.int32'>
             <class 'numpy.int64'>
             <class 'numpy.longlong'>
             <class 'numpy.timedelta64'>
         <class 'numpy.unsignedinteger'>
             <class 'numpy.uint8'>
             <class 'numpy.uint16'>
             <class 'numpy.uint32'>
             <class 'numpy.uint64'>
             <class 'numpy.ulonglong'>
     <class 'numpy.inexact'>
         <class 'numpy.floating'>
             <class 'numpy.float16'>
             <class 'numpy.float32'>
             <class 'numpy.float64'>
             <class 'numpy.longdouble'>
         <class 'numpy.complexfloating'>
             <class 'numpy.complex64'>
             <class 'numpy.complex128'>
             <class 'numpy.clongdouble'>
 <class 'numpy.flexible'>
     <class 'numpy.character'>
         <class 'numpy.bytes_'>
         <class 'numpy.str_'>
     <class 'numpy.void'>
         <class 'numpy.record'>
 <class 'numpy.bool_'>
 <class 'numpy.datetime64'>
 <class 'numpy.object_'>

```

packages/dara-core/dara/core/main.py
```diff
@@ -228,6 +230,14 @@ async def catchall_middleware(req: Request, call_next):
     for kind, handler in config.ws_handlers.items():
         custom_ws_handlers_registry.register(kind, handler)
 
+    # update encoder registry
+    for typ, encoder in config.encoders.items():
+        encoder_registry[typ] = encoder
```
Roman-Kornev, line 235: Just use .update(). It's faster than a loop.
```suggestion
encoder_registry.update(config.encoders)
```

packages/dara-core/dara/core/interactivity/derived_variable.py
```diff
@@ -204,10 +205,12 @@ def _restore_pydantic_models(func: Callable[..., Any], *args):
         if len(var_arg_idx) == 0:
             for param, arg in zip(parameters, args):
                 typ = param.annotation
-                if typ is not None and isclass(typ) and issubclass(typ, BaseModel) and arg is not None:
+                if typ in encoder_registry:
+                    parsed_args.append(encoder_registry[typ]['deserialize'](arg))
+                elif typ is not None and isclass(typ) and issubclass(typ, BaseModel) and arg is not None:
+                    parsed_args.append(typ(**arg))
```
Roman-Kornev, line 211: This branch is now identical to the one below. Do we still need it?

Roman-Kornev (approved):
Hey Aida, looks good. A couple of small improvements:

packages/dara-core/dara/core/interactivity/derived_variable.py
```diff
@@ -204,10 +205,12 @@ def _restore_pydantic_models(func: Callable[..., Any], *args):
...
+                else:
                     parsed_args.append(typ(**arg))
```
krzysztof-causalens, line 213: This should be just

```suggestion
                    parsed_args.append(arg)
```

packages/dara-core/dara/core/main.py
AidaZheng: Thanks Roman, will address that. May I ask why the dict.update() is faster?

krzysztof-causalens: I would guess it's simply because Python can internally optimize the method as compared to a manual loop.

docs/best-practices/retaining-correct-variable-types.md
```diff
@@ -1,6 +1,40 @@
 ---
 title: Retaining Correct Variable Types
 ---
+### Adding Type Annotations to Your Resolver
+`dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. When a `Variable` is passed to one of these, it is serialized into a format that FastApi can handle and becomes a python basic type format on the server side. Eg, a  `numpy.array` will became a  `list`, a `numpy.int8` will become a `int`.
```
krzysztof-causalens, line 5: `FastApi` -> `FastAPI`

Same throughout the document

krzysztof-causalens, line 5: `python` -> `Python`

docs/best-practices/retaining-correct-variable-types.md
```diff
@@ -1,6 +1,40 @@
...
+
+
+By default, Dara supports to serialize all the generic data types in Numpy. See [default data types supported](https://github.com/causalens/dara/blob/master/packages/dara-core/dara/core/internal/encoder_registry.py).
+
+You can also add your custom encoder by using `ConfigurationBuilder.add_encoder()`. Notice, if a Variable is a type that can not be serialized by either default encoder handler or custom encoder handle, it will case a crush.
```
krzysztof-causalens, line 10: > ...custom encoder handle, it will case a crush.

 

...custom encoder handle**r**, it can cause serialization to fail

docs/best-practices/retaining-correct-variable-types.md
krzysztof-causalens, line 5: Technically the serialization in `Variable` happens when the `Variable` is included anywhere in your template, not just in a DV or py_component, e.g.

`Text(text=Variable('some text'))` -> Dara will attempt to serialize the variable alongside the default.

Maybe worth simplifying this part a little bit. Here's what I'd suggest:


> `dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. This means that there are times when Dara needs to be able to serialize and deserialize your values.
> 
> Static values or `Variable` default values need to be serialized to JSON as part of the component tree - either when included directly in the page or when returned from a `py_component`.
> 
> When a user-defined resolver (for `DerivedVariable`s, `py_component`s or action) is called, Dara receives the `Variable` values from the frontend as the serialized JSON. This means your data structures would be turned into primitive types, a `numpy.array` will became a `list`, a `numpy.int8` will become a `int`. 
> 
> To preserve your original types, Dara provides an encoder system which will attempt to turn the primitive types back into the original type, as long as you annotate your resolver with a type for which an encoder is defined.

docs/best-practices/retaining-correct-variable-types.md
```diff
@@ -1,6 +1,40 @@
 ---
 title: Retaining Correct Variable Types
 ---
+### Adding Type Annotations to Your Resolver
+`dara.core.interactivity.plain_variable.Variable`s are stored in the browser and `dara.core.interactivity.derived_variable.DerivedVariable`s and `py_component`s are calculated on the server. When a `Variable` is passed to one of these, it is serialized into a format that FastApi can handle and becomes a python basic type format on the server side. Eg, a  `numpy.array` will became a  `list`, a `numpy.int8` will become a `int`.
+
+
+By default, Dara supports to serialize all the generic data types in Numpy. See [default data types supported](https://github.com/causalens/dara/blob/master/packages/dara-core/dara/core/internal/encoder_registry.py).
```
krzysztof-causalens, line 8: > By default, Dara supports to serialize all the generic data types in Numpy.

Out of the box, Dara comes with encoders for all generic data types in `pandas` and `numpy`. See...

docs/best-practices/retaining-correct-variable-types.md
```diff
@@ -1,6 +1,40 @@
...
+
+my_der_var = DerivedVariable(dv_func, variables=[my_var])
+
+# Same for py_component
+@py_component
+def my_component(var: numpy.array):
+    return Text(var.tobytes())
+my_component(my_var)
+```
+Currently the serialize/deserialize handler does not work for action resolvers due to their different API shape, which we'll hopefully solve in the near future.
```
krzysztof-causalens, line 37: > , which we'll hopefully solve in the near future.

Can remove that for now, not relevant to users. We'll update this section once we have a solution to this problem.

krzysztof-causalens, line 37: It's just the automatic deserialization which does not work. I.e., action resolvers will receive the primitive types.

krzysztof-causalens (commented):
Looks good, some stylistic comments on the docs page

docs/best-practices/retaining-correct-variable-types.md
tamara-causalens: You can say: "Current limitation of the serialize/deserialize handler is that it does not work for action resolvers due to their different API shape."

tamara-causalens: or something along those lines

tamara-causalens (approved):
Great job!

krzysztof-causalens (approved):

PR #84, opened 2023-09-26
Fix: DO-1862 custsup table search does not show up

patricia-causalens:
## Motivation and Context

Even with `searchable` set to True the search box did not appear over the Table component.

## Implementation Description

The issue was found due to Table's inset to be defaulted to zero. That meant that the Table was on top of the search bar. The inset is needed as a hack around some resizing issues we had in the past however it did not account for the bar being present. To fix this I have added a condition on searchable set, where if true adds some space for it.

Table unlike other components also had some padding set around it, this has been removed.

None

## How Has This Been Tested?

Locally by rendering a Table with and without searchable flag
On safari, firefox and chrome

## PR Checklist:

table scroll

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -1,7 +1,9 @@
 ---
 title: Changelog
 ---
+## NETX
 
+-   Fixed an issue where `Table` search bar was hidden by the table itself
 ## 1.1.8
 
 -   Fixed an issue where `Table` did not return the correct index row when sorted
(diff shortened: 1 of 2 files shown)


packages/dara-components/changelog.md
```diff
@@ -1,7 +1,9 @@
 ---
 title: Changelog
 ---
+## NETX
```
tamara-causalens, line 4: next

tamara-causalens (approved):

packages/dara-components/changelog.md
patricia-causalens: Oops, fixed

Roman-Kornev (approved):

krzysztof-causalens (approved):

PR #87, opened 2023-09-29
Feat: DO-436 introduce more cache control customization

krzysztof-causalens:
## Motivation and Context

This PR brings improvements to the existing caching functionality for server-driven variables (data-, derived-, DerivedData-). Previously users could only specify a `cache_type`, controlling the scope of the cache, i.e. whether there is a single global store, or a store per session/user for the given variable results. 

Now, the `cache` prop for variables also accepts a policy object - the change is backwards-compatible, users passing in a cache type will default to creating a default cache policy with the specified cache type. Not specifying the cache type at all will also make the variable use the new default - an LRU cache. This means apps will no longer have a built-in memory leak and entries for variables will be cleaned up unless users opt out by specifying a different policy.

See the new documentation page for details on how the API looks like from user POV.

## Implementation Description

The previous store implementation in `dara.core.internal.store.Store` is retained as it is relied upon in few user-land solutions for a simple cache-scope-aware key-value solution. In the future we can provide a better solution building upon the new primitives introduced. For now the old store API was retained, I only removed the size tracking from it as it is now included in the CacheStore.

Introduced a new `CacheStore` class, which provides a multi-layered cache structure, more tailored to our needs.
 
Screenshot 2023-09-29 at 18 03 45

The structure is presented on the diagram above:
1. `CacheStore` links particular registry entries (which correspond to e.g. a particular variable) to its own `CacheScopeStore`. 
2. Each `CacheScopeStore` links cache scopes (user IDs for user level caching, session IDs for session level caching, one global scope for global caching) with their own cache implementations. The cache implementation depends on the policy configured on the variable instance.
3. Each cache implementation maintains its entries according to the implementation and configured policy.

The implementations provided are as follows:
1. LRU cache maintains a key->value map of nodes, which also form a doubly linked-list. This means reads and writes are O(1) since we're simply reading/adding to a dict, and reconnecting nodes as needed. When an entry is read, it's simply moved to the front of the linked list. That makes the cache eviction as simple as removing elements from the rear of the list.
2. TTL cache maintains a heap to track expiration times of cache entries. The heap keeps keeps oldest entries in front of the heap. On each read or write from the cache we simply look at the front of the list and pop entries which are expired.
3. Keep-all cache is a simple dict under the hood.

In addition the cache implementations implement `pinning`. This is required because in certain cases Dara puts something into cache with the expectation that it will be able to read back from it right after - an example is data variable counts, or task results. For those purposes the caches `pin` those special entries meaning they won't be cleaned up even if the cache eviction rules are violated. Then the entries are unpinned when read back from the cache, so they will be cleaned up on the next cleanup run.

Note that this means if a pinned entry is added and for some reason it is not read back, it will stay in the cache forever. This is a potential future improvement, perhaps those pinned entries should inherently have a TTL of few minutes to make sure they are not hogging memory. This should be a rare occurrence so kept it outside the scope of this ticket.

Dev dependency for `freezegun`, a library for mocking timers to test TTL caching. 

## How Has This Been Tested?

Added unit tests for cache implementations.
Test suites for the rest of the frameworks passing.

Did some manual testing on the cache implementations.

## PR Checklist:


sam-causalens (approved):
V. cool. LGTM

patricia-causalens (approved):
LGTM, very nice! Particularly liked the way you explained the different types with examples in the docs :)

PR #95, opened 2023-10-03
Fix: DO-1899 data variable worker thread & build 1.2.1

krzysztof-causalens:
## Motivation and Context

The issue was found when upgrading another package to version 1.2.0 with the new cache changes. It was found to be because of the API used to update the store in the DataVariable constructor. 

## Implementation Description

The Store update API is asynchronous but it needs to be called within the synchronous constructor.

Previously we simply used the raw `asyncio.create_task` API to spawn a raw task as it's a quick one-off update task. This however does not work when ran within a synchronous user-defined handler (e.g. a sync py_component, action, derived variable). This is because synchronous handlers are ran using the `run_in_threadpool` starlette function, which uses `anyio.to_thread.run_sync`. This runs the function in a separate worker thread which does not have direct access to the same event loop, so `create_task` fails. 

The fix was to catch the `RuntimeError` raised when the above happens, and instead use the `anyio` API meant to run async tasks within a worker thread.

 
## PR Checklist:

- [ ] I have implemented all requirements? (see JIRA, project documentation).
- [ ] I am not affecting someone else's work, If I am, they are included as a reviewer.

### DIFF
--- a/borg.toml
+++ b/borg.toml
@@ -1,1 +1,1 @@
-version = "1.2.0"
+version = "1.2.1"
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "npmClient": "pnpm",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "packages": [
     "packages/*"
   ],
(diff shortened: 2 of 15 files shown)


tamara-causalens (approved):
Very nice!

PR #97, opened 2023-10-04
Fix: DO-1902 fix default pandas.Timestamp encoder & build 1.2.2

krzysztof-causalens:
## Motivation and Context

Issue found when testing out new 1.2 changes, the default encoder for pandas.Timestamp was previously outputting a datetime object which was unserializable. Fixed by using the same isoformat method as the native datetime object uses.

## How Has This Been Tested?

Tested by modifying code directly in the project where the issue was spotted

## PR Checklist:

### DIFF
--- a/borg.toml
+++ b/borg.toml
@@ -1,1 +1,1 @@
-version = "1.2.1"
+version = "1.2.2"
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "npmClient": "pnpm",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "packages": [
     "packages/*"
   ],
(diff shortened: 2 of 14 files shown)


tamara-causalens (approved):

patricia-causalens (approved):
Good find, how did this came up out of curiosity

krzysztof-causalens:
Added a test for all the types, fixed some other issues uncovered in the process. @tamara-causalens could you have a look if the serialization for the pandas/numpy types makes sense? For few which I couldn't figure out a generic way to deserialize I just default to raising NotImplementedError, deserialization handler only comes into play when user annotates their function with that type so in those cases they can just not annotate it and parse it themselves.

claudiawestby:
> Good find, how did this came up out of curiosity

testing out some hlcs I had a `DerivedVariable` that was of type `List[ScenarioResult]` and I noticed that it could no longer be deserialized

claudiawestby:
thanks for debugging this!

tamara-causalens (approved):

PR #98, opened 2023-10-06
Feat: DO-1888 allow implementation substitution on registry entries

krzysztof-causalens:
## Motivation and Context

In certain advanced use cases it's desirable to override the implementation of one of the core handlers on a per-registry-entry level, e.g. for certain components. 
This PR makes it possible by updating registry entries to hold references to their respective handlers implementation functions.

## Implementation Description

Registry entries for DerivedVariable, DerivedDataVariable, DataVariable, upload resolvers and action resolvers now all have extra keys which refer to their handler implementations, e.g. action resolvers have an `execute_action` key which by default points to the default implementation of action execution.

This allows the RegistryLookup to output modified entries with custom implementations. Some internals had to be adjusted to fully support this use case. In particular the data variable endpoint was adjusted to support the handler returning already-serialized data to prevent double encoding a dataframe.

Note that this comes with a typing downsides as I kept the references loosely types to prevent a bunch of circular imports for types and re-defining the handler types in the place of registry entries. The handler keys on entries do mention the reference implementation. 

## How Has This Been Tested?

All tests passing. Tested in a local project utilizing the custom implementations.

## PR Checklist:

### DIFF
--- a/packages/dara-components/dara/components/common/dropzone.py
+++ b/packages/dara-components/dara/components/common/dropzone.py
@@ -20,7 +20,7 @@ from uuid import uuid4
 
 from pandas import DataFrame
 
-from dara.core.base_definitions import Action
+from dara.core.base_definitions import Action, UploadResolverDef
 from dara.core.definitions import StyledComponentInstance
 from dara.core.interactivity import DataVariable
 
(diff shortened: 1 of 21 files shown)


tamara-causalens (approved):
Looks good! Awesome!

sam-causalens (approved):
LGTM, is this something we need to document anywhere? I'm not sure whether we documented the registry merging anywhere or not. If not then we can skip for now.

krzysztof-causalens:
> LGTM, is this something we need to document anywhere? I'm not sure whether we documented the registry merging anywhere or not. If not then we can skip for now.

Good call, but yeah for now we keep it as an internal changelog entry. Once we have it stabilized and we're confident this meets needs our of customizing the behaviour longer term, we can chuck it in an advanced docs section.

PR #104, opened 2023-10-09
Fix: DO-1913: table has a weird scrollbar in dara 1 2

patricia-causalens:
## Motivation and Context

After 1.2 release a horizontal scrollbar would always appear in Table component

## Implementation Description

This seems to be an issue with the AutoSizer calculating a width just slightly larger than it should.  This does not seem to happen if the component has left/right paddings above a certain threshold. While playing with it I noticed we could alternative set the width to 100% - 1px, it seems that in some cases defining a width to a parent component helps the AutoSizer. This seems to allow the AutoSizer to get the size just right. 

## How Has This Been Tested?

I tested locally with a layered structure of tables. In Chrome, Safari and Firefox on different screen sizes

## PR Checklist:

Screenshot 2023-10-09 at 16 42 42

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fixed an issue where `Table` would always overflow
+
 ## 1.1.10
 
 -   Internal: `parseLayoutDefinition` and `GraphLayoutDefinition` are now exposed on js side `CausalGraph` object can now accept extras.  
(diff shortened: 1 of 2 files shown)


krzysztof-causalens (approved):
Good find! Hacky solution but happy to proceed since you've tested that it works in all major browsers. Have tried that it works in all screen sizes?

Worth noting that 1.3.0 has already been released, so if the fix is supposed to be included in 2023.10 we will need to cherry pick to a 1.2.x branch and release from there. FYI @tamara-causalens

tamara-causalens:
Yep - this needs to be cherry picked to 1.2! 

btw, looks good to me!

tamara-causalens (approved):

patricia-causalens:
> Good find! Hacky solution but happy to proceed since you've tested that it works in all major browsers. Have tried that it works in all screen sizes?
> 
> Worth noting that 1.3.0 has already been released, so if the fix is supposed to be included in 2023.10 we will need to cherry pick to a 1.2.x branch and release from there. FYI @tamara-causalens

Yes I tested on laptop screen, monitor and resizing/zooming in and out of the page

PR #105, opened 2023-10-09
Docs: DO-1842: plotly error value error

patricia-causalens:
## Motivation and Context

This issue was found by an user when building an app and some Plotly figures would not render

## Implementation Description

I was unable to recreate the issue, this seems to come and go quite inconsistently. For now I have added an entry to the FAQ section of what they can do when encountering it in the future. And if someone does come across this hopefully we can get an example to try and apply a more permanent fix to. 

None

## PR Checklist:


tamara-causalens (approved):

claudiawestby (commented):
so there's actually two simple workarounds and I think it's worth just including them (while keeping the link for reference)

the first one is 
```
try:
    fig = px.scatter(...)
except Exception:
    fig = px.scatter(...)  # Exact same line as previously
return fig
```
and the second one is just instantiang using `go.Figure` and setting the layout explicitly:
```
fig = go.Figure(layout=dict(template='plotly'))
```

krzysztof-causalens (approved):

PR #109, opened 2023-10-11
Fix: DO-1953 numeric input breaks when emptied

patricia-causalens:
## Motivation and Context

It was found that when Input of type number was emptied and the user did not enter another number quickly it would crash.

## Implementation Description

The issue was due to when the NumericInput emptied its value became NaN which was not handled resulting in infinite updates to the point the component crashed. The fix was to handle it by setting to null when empty

## How Has This Been Tested?

Locally on a test version of an app

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -5,6 +5,7 @@ title: Changelog
 ## NEXT
 
 -   Fixed an issue where `Table` would always overflow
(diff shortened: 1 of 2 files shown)


tamara-causalens (approved):

claudiawestby (approved):
thank you!

krzysztof-causalens (approved):

PR #113, opened 2023-10-12
Feat: DO-1936 actions API v2

krzysztof-causalens:
## Motivation and Context

This PR implements the new action API based on the `@action` decorator. Actions are now decorated functions, similar to `@py_component`s. 
More details in the updated actions documentation page.

This API brings more control over the API with control flow, error handling etc. as well as performance improvements.

The API also simplifies and makes the internals more consistent. e.g. previously only select actions supported tasks, each one did error handling separately etc. Errors now consistently simply send a Notify action with the error to the frontend.

## Implementation Description

This change is fully backwards compatible, existing imports and structure retained their APIs. The previous APIs are now deprecated, emitting warnings when used. 

Under the hood passing resolvers to deprecated APIs creates an `@action`-decorated function.

The new useAction hook should be much more performant as it is no longer dynamically importing hooks on each render. Actions are now pure functions, so they can be globally cached.

Some of the actions are now aliases of other ones, e.g. Logout = Navigate('/logout'), DownloadFile also uses Navigate to a url with the code embedded in it rather than having separate implementation.

File download implementation was updated to use the cache store and TTL cache for code cleanup rather than a custom JWT-based solution.

Other changes:
- upgraded pylint from `^2.6.0` to `^3.0.0`
- upgraded mypy from `^0.991` to `^1.6`

Removed `pylint_print` plugin (as it was incompatible with pylint 3) in favour of using the builtin `bad_builtin` plugin, marking `print` as a disallowed builtin.

## How Has This Been Tested?

Manually:
- locally testing each API
- linking a local repo of HLCs with the new Dara version and making sure it works without changes

E2E - existing Cypress test suite (with old action APIs) passing

Unit tests - backend and frontend test suites updated to match the new APIs

## PR Checklist:


docs/getting-started/actions.md
```diff
@@ -2,73 +2,143 @@
...
+:::note
+
+The action methods on `ActionCtx` are `async` and must be `await`-ed, so your decorated function should be `async` as well.
 
-To see how the `UpdateVariable` works, consider the example below:
+:::
 
 ```python
-from dara.core import ConfigurationBuilder, Variable, UpdateVariable
+from dara.core import action, Variable
```
tamara-causalens, line 27: Can we try to make this example work end to end so that it allows the reader to just c&p and run? Can be a simple example which demonstrates the main point

tamara-causalens: sorry, missread

docs/getting-started/actions.md
```diff
@@ -2,73 +2,143 @@
...
-## `UpdateVariable`
+A valid argument to an action parameter is one of the following three options:
 
-The `dara.core.interactivity.actions.UpdateVariable` action can trigger the update of a `Variable`, `DataVariable` or `UrlVariable`.
+- an instance of an `@action`-decorated function (as in the example above)
+- an instance of an individual static action object
+- a list of a combination of the above two options
 
-It takes the following arguments:
+In this section we will go through all the different ways to create actions and how to use them.
```
tamara-causalens, line 13: "In this section you will learn all the different ways ..."

packages/dara-core/dara/core/interactivity/plain_variable.py
```diff
@@ -89,6 +89,85 @@ def get(self, key: str):
...
+
+        ```python
+
+        from dara.core import Variable
+        from dara.components import Select
+
+        var = Variable('first')
+
+        Select(
+            value=Variable('first'),
```
tamara-causalens, line 104: just `var` is enough here

packages/dara-core/dara/core/interactivity/url_variable.py
```diff
@@ -51,6 +51,85 @@ def __init__(self, query: str, default: Optional[VariableType] = None, uid: Opti
...
+
+        ```python
+
+        from dara.core import UrlVariable
+        from dara.components import Select
+
+        var = UrlVariable('first', query='num')
+
+        Select(
+            value=UrlVariable('first'),
```
tamara-causalens, line 66: just `var`

packages/dara-core/dara/core/interactivity/url_variable.py
```diff
@@ -51,6 +51,85 @@ def __init__(self, query: str, default: Optional[VariableType] = None, uid: Opti
...
+        ```python
+
+        from dara.core import UrlVariable
+        from dara.components import Button
+
+        show = UrlVariable(True, query='show')
+
+        Button(
+            'Hide',
+            onclick=var.update(False),
```
tamara-causalens, line 120: should this be `show`?

packages/dara-core/dara/core/visual/template.py
```diff
@@ -44,7 +44,7 @@ def add_route(
         content: ComponentInstance,
         icon: Optional[str] = None,
         include_in_menu: Optional[bool] = True,
-        reset_vars_on_load: Optional[List[AnyVariable]] = [],
+        on_load: Optional[Action] = None,
```
tamara-causalens, line 47: Is this technically a breaking change?

docs/getting-started/actions.md
```diff
@@ -176,64 +290,473 @@ config.add_page(name='Test Page', content=test_page())
...
+
+```python
+from dara.core import action, Variable
+from dara.components import Button
+
+var = Variable(default=False)
+
+@action
+async def toggle_action(ctx: action.Ctx):
+    await ctx.update(target=var, value=not ctx.input)
```
patricia-causalens, line 519: I am confused, isn't input of a Button always None?

packages/dara-core/dara/core/visual/template.py
krzysztof-causalens: Both `reset_vars_on_load` and `on_load` exist on the public API `config.add_page`. 

We have to draw the line on what's "internal" vs "public API" somewhere, and I'd consider this template API more internal. In theory since this is Python you can import any of Dara internals and use them. 

In this case, even the custom templates we use e.g. in the CLI do not use this API directly. I might be wrong though and this might be a commonly used API, happy to retain the old kwarg here if that's the case.

tamara-causalens: makes sense, I'm ok with that. Not sure it is often used

docs/getting-started/actions.md
```diff
@@ -176,64 +290,473 @@ config.add_page(name='Test Page', content=test_page())
...
+from dara.components import Input, Stack, Button, Text
+
+config = ConfigurationBuilder()
+
+my_var = Variable(1)
+der_var = DerivedVariable(lambda x: float(x) ** 2 , variables=[my_var], deps=[])
+
+@action
+async def trigger_my_var(ctx: action.Ctx):
+    await ctx.trigger(variable=my_var)
```
patricia-causalens, line 548: Should this be der_var?

sam-causalens (approved):
Nice job, the changes look good to me.

docs/getting-started/actions.md
```diff
@@ -176,64 +290,473 @@ config.add_page(name='Test Page', content=test_page())
...
+@action
+async def action1(ctx: action.Ctx, previous_value: int):
+    await ctx.update(target=var, value=1)
+
+@action
+async def action2(ctx: action.Ctx):
+    await ctx.update(target=var, value=2)
+
+# Mixing an @action-annotated action with shortcut actions/action implementation objects
+Button('composing different action types', onclick=[action1(var), var.reset()])
```
patricia-causalens, line 631: Should this be calling action2 instead of reset?

patricia-causalens (commented):
Still reviewing just pushing first checkpoint for comments :)

docs/getting-started/actions.md
```diff
@@ -2,73 +2,143 @@
...
+some_variable = Variable()
+
+@action
+async def my_action(ctx: action.Ctx, inp):
+    # Value coming from the component, in this case the selected item
+    value = ctx.input
+    # Your action logic...
+
+    # Update `some_variable` to `value` multiplied by 2
+    await ctx.update(target=some_variable, value=value * 2)
```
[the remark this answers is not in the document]
krzysztof-causalens: Yep exactly. I'll update this example to show both a static value and a variable value being passed in

docs/getting-started/actions.md
```diff
@@ -2,73 +2,143 @@
...
+async def trigger(
+    variable: DerivedVariable,
+    force: bool = True
+)
+```
+
+`trigger` is a method to force a recalculation of a `DerivedVariable`. It takes the following arguments:
 
-## `TriggerVariable`
+- `variable` - a `DerivedVariable` instance to force a recalculation of,
```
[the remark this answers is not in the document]
krzysztof-causalens: Yeah sure we could make them both consistent. Which one do you think makes more sense?

docs/getting-started/actions.md
```diff
@@ -77,97 +147,141 @@ config.add_page(name='Trigger Variable', content=test_page())
...
+```
 
-The `dara.core.interactivity.actions.NavigateTo` action will trigger a change in route based on the url passed. The url can be a static string, or it can be a function that will be called with the element that triggered the action.
+`navigate` is a method to navigate to a new URL. It takes the following arguments:
 
-In the example below, on 'Test Page', clicking on a button 'Go to Another Page' takes you to the 'Another Page' page. This demonstrates how you can pass a static url to the `NavigateTo` action. On the 'Another Page' page, you can see how to control `NavigateTo` by passing url as a function based on component value.
+- `url` - the URL to navigate to
+- `new_tab` - a boolean indicating whether to open the URL in a new tab
+
+See the following example:
```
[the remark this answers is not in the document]
krzysztof-causalens: Yeah I just felt like it's unnecessary and the page is already pretty detailed, with the same examples being reiterated with different APIs. If you feel like any of them in particular are crucial for making it easier to understand happy to bring them back

docs/getting-started/actions.md
```diff
@@ -176,64 +290,473 @@ config.add_page(name='Test Page', content=test_page())
...
+)
+```
 
-config = ConfigurationBuilder()
+`download_file` is a method to download a file from the server. It takes the following arguments:
 
-x = Variable(0)
-y = Variable(1)
-z = Variable(2)
+- `path` - the path to the file to download
```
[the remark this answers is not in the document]
krzysztof-causalens: The path is just passed almost directly into an `open` call, so should work pretty much as expected.

docs/getting-started/actions.md
```diff
@@ -176,64 +290,473 @@ config.add_page(name='Test Page', content=test_page())
...
 
 
 config.add_page(name='Download Variable', content=test_page())
 ```
 
-## `DownloadContent`
+## Shortcut actions
+
+Shortcut actions are convenience methods on variable instances that can be used to perform common actions.
+Beside being more consise and easier to read, they are also more performant as they do not require a roundtrip to the server in order
```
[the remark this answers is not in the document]
krzysztof-causalens: 🤦 good catch

docs/getting-started/actions.md
krzysztof-causalens: Oops good catch, will come up with a better example

docs/getting-started/actions.md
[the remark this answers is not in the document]
patricia-causalens: Variable seems to be in most places throughout already we could just keep with it

docs/getting-started/actions.md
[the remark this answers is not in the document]
patricia-causalens: I think most are fine, for the url example maybe worth having somewhere in the code example that you are navigating to a different page just in case we have someone new to coding/technical things and that don't have very clear what it means.

docs/getting-started/actions.md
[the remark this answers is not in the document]
krzysztof-causalens: Makes sense, will update to `variable` then

docs/getting-started/interactivity.mdx
```diff
@@ -140,48 +140,78 @@ In the above example, whenever the `Input` field changes, it updates the value o
...
+from dara.core import action, Variable
+from dara.components import Select, Item
 
-my_var = Variable(5)
+some_variable = Variable()
 
-def add_one(ctx: UpdateVariable.Ctx):
-    return ctx.inputs.old + 1
+@action
+async def my_action(ctx: action.Ctx, inp):
```
claudiawestby, line 162: what is `inp` here?

docs/getting-started/actions.md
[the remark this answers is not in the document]
krzysztof-causalens: Fair enough, added back the explanation for the navigate example

docs/getting-started/interactivity.mdx
```diff
@@ -140,48 +140,78 @@ In the above example, whenever the `Input` field changes, it updates the value o
...
 
-config.add_page(name='Increment Variable', content=test_page())
+Select(
+    items=[Item(label='item1', value=1), Item(label='item2', value=2)],
+    onchange=my_action(2)
+)
 ```
 
-The `dara.components.common.button.Button` component has a parameter `onclick` which is the component's callback. You can pass it an action, which in this case is an `UpdateVariable` instance.
+The example above shows how to use the `@action` decorator to create an action that updates a `Variable` with the value of the selected item in a `Select` component.
```
claudiawestby, line 177: I'm a bit confused here. I am guessing `value=value * 2` was supposed to be `value=value * inp` and if this is the case I would explain better what this example is doing which is rather updating a variable with the value of the selected item multiplied by 2

claudiawestby: yep looks good!

docs/getting-started/actions.md
```diff
@@ -2,73 +2,144 @@
...
-- `variable`: the `Variable`, `DataVariable` or `UrlVariable` to update with a new value upon triggering the action
-- `extras`: any extra variables to resolve and pass to the resolver function, you can obtain a list of the resolved values of all the extras passed with `ctx.extras` in the resolver function
+## `@action` decorator
+
+The primary way to create actions in Dara is the `@action` decorator. This decorator takes a function and returns an action that can be passed to a component's callback.
+It injects an `ActionCtx` object as the first argument of the function, which contains the input sent from the component and exposes action methods to update the state of the application.
+
+:::note
+
+The action methods on `ActionCtx` are `async` and must be `await`-ed, so your decorated function should be `async` as well.
```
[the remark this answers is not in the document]
krzysztof-causalens: Tbh on the surface level there isn't much you need to know to use this API - just mark your action as `async` and `await` methods which are marked `async`. I appreciate it's hard for me to put myself in a data scientist's shoes so not sure how to explain it to them well. 
Do you have a particular resource that you think would work well? I can think of either a generic introduction like this blogpost https://realpython.com/async-io-python/ or FastAPI's section https://fastapi.tiangolo.com/async/#in-a-hurry (we've linked to this in other places).

packages/dara-core/dara/core/interactivity/any_variable.py
```diff
@@ -298,6 +298,31 @@ def __le__(self, value: Any) -> Condition:
...
+        ```python
+
+        from dara.core import Variable
+        from dara.components import Button
+
+        counter = Variable(default=0)
+
+        Button(
+            'Reset',
+            onclick=var.reset(),
```
patricia-causalens, line 316: I think this should be counter.reset()

packages/dara-core/dara/core/interactivity/plain_variable.py
```diff
@@ -89,6 +89,85 @@ def get(self, key: str):
...
+
+        from dara.core import Variable
+        from dara.components import Select
+
+        var = Variable('first')
+
+        Select(
+            value=var,
+            items=['first', 'second', 'third'],
+            onchange=var.sync(),
```
patricia-causalens, line 106: var is already synced with the component as it is passed as value would it make more sense if we had a second var in this case? Same with UrlVariable

packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -101,6 +105,27 @@ def _get_pandas_array_encoder(array_type: Type[Any], dtype: Any, raise_: bool =
...
+
+    Looks up the type in the encoder_registry and uses the deserializer to convert the value into the given type.
+
+    :param value: the value to deserialize
+    :param typ: the type to deserialize into
+    """
+    if isinstance(value, BaseTask):
+        return value
+
+    if typ is not None and typ in encoder_registry:
```
patricia-causalens, line 125: more of a general question, if an if condition is used more than one time do we prefer having it repeated and larger statements or nested ifs?

docs/getting-started/actions.md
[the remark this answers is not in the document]
claudiawestby: agreed that they don't need to know much and I think that's a good thing! but I can see it peaking their curiosity as to how things like that work and I don't want to give the impression that we just expect them to be familiar with it. I like the first link as it puts things into context of other concurrency and paralellism techniques which a data scientist probably will be familiar with.

_however_ if you think it might be distracting and lead the reader astray during an important part of the docs then I could be persuaded to not link anything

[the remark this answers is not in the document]
krzysztof-causalens: Fair enough, thanks. This is just a 'note' section so shouldn't be too distracting, will link the blogpost then.

claudiawestby (approved):
docs are really straightforward and comprehensive, excited to use the new API!

patricia-causalens (changes requested):
Great work Krzys! Just some minor comments, also I may have missed it but are there any tests for the new action shortcuts on the python side?

packages/dara-core/dara/core/interactivity/plain_variable.py
krzysztof-causalens: Fair point, will add a second variable

krzysztof-causalens:
@patricia-causalens good catch, shortcuts were a late addition so didn't test them too much. Will add few quick tests just to cover those shortcuts.

patricia-causalens (approved):
LGTM

PR #120, opened 2023-10-17
Fix: DO-1979 DataFrame serialization to dict

AidaZheng:
## Motivation and Context
DataFrame should be serialised to dict rather than json string

No

## How Has This Been Tested?
unit tested

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -16,7 +16,7 @@ limitations under the License.
 """
 import json
 from collections.abc import Mapping
-
+from fastapi.encoders import jsonable_encoder
 # pylint: disable=unnecessary-lambda
 from typing import Any, Callable, MutableMapping, Type
 
@@ -118,6 +118,6 @@ encoder_registry: MutableMapping[Type[Any], Encoder] = {
     pandas.Index: Encoder(serialize=lambda x: x.to_list(), deserialize=lambda x: pandas.Index(x)),
     pandas.Timestamp: Encoder(serialize=lambda x: x.isoformat(), deserialize=lambda x: pandas.Timestamp(x)),
     pandas.DataFrame: Encoder(
-        serialize=lambda x: x.to_dict(orient='records'), deserialize=lambda x: _df_decode_resolver(x)
-    ),
+        serialize=lambda x: jsonable_encoder(x.to_dict(orient='dict')), deserialize=lambda x: _not_implemented(x,pandas.DataFrame)
+    )
 }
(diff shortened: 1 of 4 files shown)


packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -114,5 +117,7 @@ def _df_decode_resolver(df: Any):
     pandas.Series: Encoder(serialize=lambda x: x.to_list(), deserialize=lambda x: pandas.Series(x)),
     pandas.Index: Encoder(serialize=lambda x: x.to_list(), deserialize=lambda x: pandas.Index(x)),
     pandas.Timestamp: Encoder(serialize=lambda x: x.isoformat(), deserialize=lambda x: pandas.Timestamp(x)),
-    pandas.DataFrame: Encoder(serialize=lambda x: x.to_json(orient='records'), deserialize=lambda x: _df_decode_resolver(x)),
+    pandas.DataFrame: Encoder(
+        serialize=lambda x: x.to_dict(orient='records'), deserialize=lambda x: _df_decode_resolver(x)
```
krzysztof-causalens, line 121: Won't this output an unserializable dict if e.g. the index is using Timestamp objects? If that's the case perhaps we need to call `jsonable_encoder` on the output of `to_dict` here.

krzysztof-causalens, line 121: Is orient='record' the one users would usually expect? cc @claudiawestby

claudiawestby: using records will eliminate the index which I would advise against. this is really just a matter of preference, at causalens we use `orient="list"` a lot because that's what `CausalNet` _used_ to expect but with 3.0 refactors it's no longer. `orient="list"` also removes the index so I wouldn't go with that one anyways

the default is `orient="dict"` which does keep the index so I would suggest sticking with that

packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -99,6 +100,7 @@ def _get_pandas_array_encoder(array_type: Type[Any], dtype: Any, raise_: bool =
     pandas.Index: Encoder(serialize=lambda x: x.to_list(), deserialize=lambda x: pandas.Index(x)),
     pandas.Timestamp: Encoder(serialize=lambda x: x.isoformat(), deserialize=lambda x: pandas.Timestamp(x)),
     pandas.DataFrame: Encoder(
-        serialize=lambda x: jsonable_encoder(x.to_dict(orient='dict')), deserialize=lambda x: _not_implemented(x,pandas.DataFrame)
-    )
+        serialize=lambda x: jsonable_encoder(x.to_dict(orient='dict')),
+        deserialize=lambda x: _not_implemented(x, pandas.DataFrame),
```
krzysztof-causalens, line 104: Actually, could you test how this behaves with DataVariables? i.e. you should still be able to annotate an input as DataFrame for data variables:

```python
def resolver(x: pandas.DataFrame):
  foo

data_var = DataVariable(...)

DerivedVariable(resolver, variables=[data_var])
```

This might attempt to run through the deserialize logic? e.g. the deserializer should work if `x` is already a DataFrame. Otherwise it should fail as we can't know exactly how to process a dict into a DataFrame, preserving the types correctly

krzysztof-causalens (approved):
LGTM, nice catch on the task issue. Good to merge from my side if @claudiawestby is happy with the serialization method

claudiawestby (approved):
looks good from what I can see 👍 happy to test on my scenario hlc after it's released

PR #123, opened 2023-10-19
Improvement: DO-1993: Add explicit value support for Select/Multiselect/Combobox

Roman-Kornev:
## Motivation and Context
Currently, the select component will not display a value if it's not also present in the list of items. This presents a problem when that list of items uses a piece of derived state, which might not be available at the time of render.

This PR lets us explicitly state a `value` of a Select/Multiselect/Combobox, even if it's not in the list of `items`.

## Implementation Description
If an explicit value is set, it can either be a primitive, in which case it will get converted to an `Item`, or an object, which will be displayed as is.

For the multiselect case, every primitive item in the array will be converted to an `Item` as well.

## How Has This Been Tested?
Locally

## PR Checklist:

image

### DIFF
--- a/packages/dara-components/js/common/select/select.tsx
+++ b/packages/dara-components/js/common/select/select.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable react-hooks/exhaustive-deps */
 
+import { isArray } from 'lodash';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import {
@@ -36,6 +37,9 @@ export function getMultiselectItems(values: Array<any>, items: Array<Item>): Arr
         return acc;
     }, []);
 }
+
+const toItem = (value: any): Item | undefined | null =>
+    typeof value === 'string' || typeof value === 'number' ? { label: String(value), value } : value;
 
 const StyledSelect = injectCss(UiSelect);
 const StyledMultiSelect = injectCss(MultiSelect);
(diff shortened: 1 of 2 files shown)


tamara-causalens (approved):

Roman-Kornev:
> If someone had a list for a select and the value is updated from outside. But they made a mistake and the value is not in the list, maybe a typo. What would happen?

Previously, the value displayed would disappear. Now it will stay, as the value is present

Roman-Kornev:
Just had a chat with Patricia to help explain the issue.
I've added a changelog entry, should be good to go.

packages/dara-components/js/common/select/select.tsx
```diff
@@ -169,7 +173,10 @@ function Select(props: SelectProps): JSX.Element {
             />
         );
     }
-    const [selectedItem, setSelectedItem] = useState(itemArray.find((item) => String(item.value) === String(value)));
+    const explicitValue = toItem(value);
+    const [selectedItem, setSelectedItem] = useState(
+        explicitValue ?? itemArray.find((item) => String(item.value) === String(value))
```
krzysztof-causalens, line 178: Should this condition not be reversed? i.e. use the item from the list and fall back to the explicit one

packages/dara-components/js/common/select/select.tsx
```diff
@@ -137,9 +141,9 @@ function Select(props: SelectProps): JSX.Element {
 
     // We need to redefine items as the type is not known at this point
     const itemArray = formattedItems as Array<Item>;
-
     if (props.multiselect) {
-        const [selectedItems, setSelectedItems] = useState(getMultiselectItems(value, itemArray));
+        const explicitValues = isArray(value) ? value.map(toItem) : value;
+        const [selectedItems, setSelectedItems] = useState(explicitValues ?? getMultiselectItems(value, itemArray));
```
krzysztof-causalens, line 146: Same here, see other comment

krzysztof-causalens (approved):
Looks good, one quick question

PR #126, opened 2023-10-20
Fix: DO-2004 deserializer running for already serialized values

krzysztof-causalens:
## Motivation and Context

Previous PR moved the shared arg restoration logic to a central `deserialize` method (previously scattered in DV and py_component logic) as it's now also reused by actions.

It turns out there was a bug there where when the incoming type was already correct (e.g. the type was coming from a derived variable - serverside - or a static kwarg) the deserializer would still attempt to run and fail in some cases.

## Implementation Description

Several improvements made to the deserializer:
- ensured if `type(value) == type`, skip deserialization
- handle `Optional[foo]` to call `deserialize(value, foo)`
- on failed deserialization, the raw value is returned and an error is logged; this means in case of an error it should still be clear what went wrong and we display an error message

## How Has This Been Tested?

Tested locally on a demo app where the issue occured, copying over the fixed file resolved the issue.
Updated tests to cover the restoration logic for actions, py_components and DVs
Added tests for the deserialize method to cover all cases

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fix an issue where argument restoration for actions, derived variables and `py_component`s would attempt to restore a value to the annotated type even when the value was already of the correct type
+
 ## 1.4.0
 
 -   Implement new action API in the form of the `@action` decorator. This decorator takes a function and returns an action that can be passed to a component's callback. It injects an `ActionCtx` object (aliased as `action.Ctx`) as the first argument of the function, which contains the input sent from the component and exposes action methods. This allows for full control over the action's behaviour, including the ability to conditionally execute specific actions with control flow, error handling etc. See the updated `actions` documentation page for more details on the new API and migration guide for the existing APIs.
--- a/packages/dara-core/dara/core/interactivity/derived_variable.py
+++ b/packages/dara-core/dara/core/interactivity/derived_variable.py
@@ -211,6 +211,12 @@ class DerivedVariable(NonDataVariable, Generic[VariableType]):
 
     @staticmethod
     def _restore_pydantic_models(func: Callable[..., Any], *args):
+        """
+        Restore argument types based on their annotations.
+
+        :param func: the function to restore the arguments for
+        :param args: the arguments to restore
+        """
         parsed_args = []
         parameters = list(signature(func).parameters.values())
         # Scan the list for any var arg or kwarg arguments
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -26,7 +26,6 @@ from pydantic import BaseModel
 from typing_extensions import TypedDict
 
 from dara.core.base_definitions import BaseTask
-from dara.core.logging import dev_logger
 
 
 class Encoder(TypedDict):
@@ -123,7 +122,7 @@ def deserialize(value: Any, typ: Optional[Type]):
     :param value: the value to deserialize
     :param typ: the type to deserialize into
     """
-    # This funtion is commonly used to deserialize parameters, which can be empty rather than None
+    # This funtion is commonly used to deserialize parameters, which can be Parameter.empty rather than None
     if typ == Parameter.empty:
         return value
 
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -155,7 +155,7 @@ def deserialize(value: Any, typ: Optional[Type]):
         if isclass(typ) and issubclass(typ, BaseModel) and value is not None:
             return typ(**value)
     except Exception as e:
-        dev_logger.error(f'Failed to deserialize value {value} into type {typ}', e)
+        dev_logger.error(f'Failed to deserialize value {value} into expected type {typ}. Consider defining a custom deserializer for the type using the `config.add_encoder` API', e)
 
     # Fall back to returning the value
     return value
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -155,7 +155,7 @@ def deserialize(value: Any, typ: Optional[Type]):
         if isclass(typ) and issubclass(typ, BaseModel) and value is not None:
             return typ(**value)
     except Exception as e:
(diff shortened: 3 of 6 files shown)


krzysztof-causalens:
Open question - should we log an error if deserialization happens and return the raw value, or just let the error happen? I'm worried the behaviour currently implemented in the PR might result in two errors being shown - one from us catching the invalid deserialization, another one from user's logic expecting a certain type. 
@causalens/decisionapps-team I'd be keen to hear others' opinions on the matter

patricia-causalens:
> Open question - should we log an error if deserialization happens and return the raw value, or just let the error happen? I'm worried the behaviour currently implemented in the PR might result in two errors being shown - one from us catching the invalid deserialization, another one from user's logic expecting a certain type. @causalens/decisionapps-team I'd be keen to hear others' opinions on the matter

If we don't log are there any cases where it should have been an error and it is not shown to the user?

krzysztof-causalens:
> > Open question - should we log an error if deserialization happens and return the raw value, or just let the error happen? I'm worried the behaviour currently implemented in the PR might result in two errors being shown - one from us catching the invalid deserialization, another one from user's logic expecting a certain type. @causalens/decisionapps-team I'd be keen to hear others' opinions on the matter
> 
> 
> 
> If we don't log are there any cases where it should have been an error and it is not shown to the user?

Sorry if I wasn't clear. If we don't explicitly catch and ignore the error you would still see it - the app would crash with the error as it currently does. The benefit is that you don't run into the potential issue I described above, where your code is written to expect the annotated type so you can run into other errors making it harder to debug. If we just crash on deserialization you will only see the deserialization error.

claudiawestby (approved):

patricia-causalens:
> Sorry if I wasn't clear. If we don't explicitly catch and ignore the error you would still see it - the app would crash with the error as it currently does. The benefit is that you don't run into the potential issue I described above, where your code is written to expect the annotated type so you can run into other errors making it harder to debug. If we just crash on deserialization you will only see the deserialization error.

Sorry still trying to understand what you mean, is it:
Scenario 1: If there is an error during deserialisation the app crashes, so the user knows immediately that something went wrong.
Scenario 2: If an error happens we catch it and log it, but the app will not crash, instead we return the raw value. Which may or may not cause lead to errors if the user is expecting it on a given format

krzysztof-causalens:
@patricia-causalens  Let me show an example, imagine this is your code:
```python
from pydantic import BaseModel

class MyModel(BaseModel):
 value: str

 # custom init 
 def __init__(self, other_value: str):
   super().__init__(value=other_value)


@py_component
def my_py_component(model: MyModel):
  return Text(model.value)

var = Variable(default=MyModel('foo'))


my_py_component(var)
```

Here `MyModel` has a custom `__init__` so our default logic for re-creating pydantic models wouldn't work, as it would call `MyModel(**{'value': 'foo'})` and `MyModel`'s `__init__` is expecting `other_value` kwarg, not `value`.
This can happen in other cases, where some other type cannot be recreated based on the serialized type, this is just one example case.

So we have two scenarios:
1. We keep the current behaviour - just let it fail. You will see an error message saying that `MyModel` expects `value `not `other_value` or something like that. We can optionally capture the error and add a helpful description of why this can happen and how to proceed. Important part is it will error on the deserialization part and will **not** run your function.
2. We capture the error, log it, but pass through a raw value and still run your function. In this case you will see two errors - one about failed deserialization, and another one because you tried to do `model.value` within your function but we gave you a `dict` instead of `MyModel` instance. My concern is this is confusing and users might not scroll past the second error to see that the root of it was actually the failed deserialization.

I'm leaning towards 1. then, where we still raise the error but we'll add a helpful error message such as:
> Failed to deserialize value {value} into expected type {typ}. Consider defining a custom deserializer for the type using the `config.add_encoder` API or removing the type annotation and handling the raw value manually.


Thoughts?

claudiawestby:
After some thought, I agree that 1. is the way to go. The more I make HLCs the more I realize it is better to tell the user what is wrong and how to fix it rather than try to work around it within the package. Anything that could lead to unexpected behaviour, the user should be very aware of in the development process.

I guess on that note, in the issue I was having in my PR, I would just remove the `Scenario` type annotation?

patricia-causalens:
> @patricia-causalens Yes, that's basically it. Let me show an example, imagine this is your code:
> 
> ```python
> from pydantic import BaseModel
> 
> class MyModel(BaseModel):
>  value: str
> 
>  # custom init 
>  def __init__(self, other_value: str):
>    super().__init__(value=other_value)
> 
> 
> @py_component
> def my_py_component(model: MyModel):
>   return Text(model.value)
> 
> var = Variable(default=MyModel('foo'))
> 
> 
> my_py_component(var)
> ```
> 
> Here `MyModel` has a custom `__init__` so our default logic for re-creating pydantic models wouldn't work, as it would call `MyModel(**{'value': 'foo'})` and `MyModel`'s `__init__` is expecting `other_value` kwarg, not `value`. This can happen in other cases, where some other type cannot be recreated based on the serialized type, this is just one example case.
> 
> So we have two scenarios:
> 
> 1. We keep the current behaviour - just let it fail. You will see an error message saying that `MyModel` expects `value `not `other_value` or something like that. We can optionally capture the error and add a helpful description of why this can happen and how to proceed. Important part is it will error on the deserialization part and will **not** run your function.
> 2. We capture the error, log it, but pass through a raw value and still run your function. In this case you will see two errors - one about failed deserialization, and another one because you tried to do `model.value` within your function but we gave you a `dict` instead of `MyModel` instance. My concern is this is confusing and users might not scroll past the second error to see that the root of it was actually the failed deserialization.
> 
> I'm leaning towards 1. then, where we still raise the error but we'll add a helpful error message such as:
> 
> > Failed to deserialize value {value} into expected type {typ}. Consider defining a custom deserializer for the type using the `config.add_encoder` API or removing the type annotation and handling the raw value manually.
> 
> Thoughts?

Thank you makes sense, yes I also prefer 1

krzysztof-causalens:
> After some thought, I agree that 1. is the way to go. The more I make HLCs the more I realize it is better to tell the user what is wrong and how to fix it rather than try to work around it within the package. Anything that could lead to unexpected behaviour, the user should be very aware of in the development process.
> 
> I guess on that note, in the issue I was having in my PR, I would just remove the `Scenario` type annotation?

If the default deserialization does not work, you either remove the annotation and get the raw result (i.e. potentially dict) or use `config.add_encoder` and handle your type explicitly.

patricia-causalens (approved):

PR #131, opened 2023-10-23
Fix: DO-2008 Add CustomResponse class

AidaZheng:
## Motivation and Context
FastApi can't serialise float('nan')/float('inf')
Hence, add a CustomResponse to handle this case when json serialisation failed

## Implementation Description
If json serialisation failed and raise a ValueError, will try to encode float('nan')/float('inf') to None and serialise again

None

## How Has This Been Tested?
Locally tested with cldp-hlc app.

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/main.py
+++ b/packages/dara-core/dara/core/main.py
@@ -14,10 +14,11 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-
 import asyncio
+import json
 import os
 import sys
+import typing
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import asynccontextmanager
 from importlib.util import find_spec
@@ -26,10 +27,12 @@ from typing import Optional
 
 from anyio import create_task_group
 from fastapi import FastAPI, HTTPException, Request
-from fastapi.responses import HTMLResponse
+from fastapi.encoders import jsonable_encoder
+from fastapi.responses import HTMLResponse, Response
 from fastapi.staticfiles import StaticFiles
 from prometheus_client import start_http_server
 from pydantic.json import ENCODERS_BY_TYPE
+from starlette.background import BackgroundTask
 from starlette.templating import Jinja2Templates, _TemplateResponse
 
 from dara.core.auth import auth_router
@@ -182,7 +185,10 @@ def _start_application(config: Configuration):
                 eng_logger.debug('Task pool shut down')
 
     app = FastAPI(
-        lifespan=lifespan, docs_url=None if is_production else '/docs', redoc_url=None if is_production else '/redoc'
+        lifespan=lifespan,
+        docs_url=None if is_production else '/docs',
+        redoc_url=None if is_production else '/redoc',
+        default_response_class=CustomResponse,
     )
 
     # Define catch-all mechanisms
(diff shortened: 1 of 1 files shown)


packages/dara-core/dara/core/main.py
```diff
@@ -423,3 +429,40 @@ def start(extra=None):
     config._run_discovery(config_module)
 
     return _start_application(config=config._to_configuration())
+
+
+class CustomResponse(Response):
```
krzysztof-causalens, line 434: Can we move it somewhere within `dara.core.internals` to not pollute the main file?

packages/dara-core/dara/core/main.py
```diff
@@ -423,3 +429,40 @@ def start(extra=None):
...
+        media_type: typing.Optional[str] = None,
+        background: typing.Optional[BackgroundTask] = None,
+    ) -> None:
+        super().__init__(content, status_code, headers, media_type, background)
+
+    def render(self, content: typing.Any) -> bytes:
+        import math
+
+        custom_encoder = {}
+        custom_encoder[float] = lambda x: None if math.isnan(x) or math.isinf(x) else x
```
krzysztof-causalens, line 451: Let's move these out of `render` to the top-level of the file, doesn't need to be redefined and reimported within `render` each time, as this is ran for every request this can affect performance.

krzysztof-causalens (commented):
Looks good, can we add a quick integration test that sends a request for e.g. a DerivedVariable which returns a DataFrame which includes some NaNs and Infs so we check that the entire serialization pipeline behaves as it should?

packages/dara-core/tests/python/test_py_component.py
```diff
@@ -266,6 +266,59 @@ def TestBasicComp(input_val: int):
...
+
+    # Test DerivedVariable can handle df with nan correctly
+    def func(var:pandas.DataFrame|None):
+        return pandas.DataFrame({'a':[1,2,numpy.nan]})
+
+    derived = DerivedVariable(func, variables=[var])
+
+    @py_component
+    def TestBasicComp(input_val: pandas.DataFrame):
+        return MockComponent(text=input_val.to_string())
```
krzysztof-causalens, line 289: Does this actually run the resolver? Here you're explicitly `to_string`-ing the DataFrame so I don't think this will run the implicit serializer

packages/dara-core/tests/python/test_py_component.py
```diff
@@ -266,6 +266,59 @@ def TestBasicComp(input_val: int):
...
+                'uid': component.get('uid'),
+                'values': {
+                    'input_val': {'type': 'derived', 'uid': str(derived.uid), 'values': pandas.DataFrame({'a':[1,2,numpy.nan]})},
+                },
+                'ws_channel': 'test_channel',
+            },
+        )
+
+        assert response.status_code == 200
+        assert response.json() == {'name': 'MockComponent', 'props': {'text': '     a\n0  1.0\n1  2.0\n2  NaN'}, 'uid': 'uid'}
```
krzysztof-causalens, line 321: The expected behaviour would be that `text` is a dict with `None`'s instead of `NaN`s.

packages/dara-core/tests/python/test_py_component.py
```diff
@@ -266,6 +266,59 @@ def TestBasicComp(input_val: int):
...
+    import pandas
+    import numpy
+
+    builder = ConfigurationBuilder()
+
+    var = Variable()
+
+    # Test DerivedVariable can handle df with nan correctly
+    def func(var:pandas.DataFrame|None):
+        return pandas.DataFrame({'a':[1,2,numpy.nan]})
```
krzysztof-causalens, line 283: Let's add an `inf` as well for good measure, and mix it with raw `float('nan')` and `float('inf')` in addition to the numpy types.

krzysztof-causalens (changes requested):

packages/dara-core/tests/python/test_variable.py
```diff
@@ -70,3 +77,31 @@ def test_derived_variable_uid_run_as_task(self):
 
         with self.assertRaises(ValueError):
             DerivedVariable(test_resolver, variables=[variable], uid='test_task_var'),
+
+async def test_derived_variables_with_df_nan():
```
krzysztof-causalens, line 81: You could have kept it as a py_component test, it was just about not calling `to_string` on the DF directly so we test the serialization. This works just as well though :)

krzysztof-causalens (approved):
LGTM

PR #132, opened 2023-10-23
Fix: DO-1984 numeric input controlled value

patricia-causalens:
## Motivation and Context

When using an input of type number in scenario manager I noticed that it would not load with any value.

## Implementation Description

The issue was found due to if the variable value became a string in the update then the component could would not show it. The fix was to handle external updates by converting it to a number, in the same way that we do with its own updates.

None

## How Has This Been Tested?

Locally by updating the variable's value through actions

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fixed issue where in some cases if `Input` of type number value variable was update outside the component, the value would not show.
+
 ## 1.4.0
 
 -   Added support for displaying a `value` in `Select` that may not be part of the `items` list.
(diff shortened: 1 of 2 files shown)


krzysztof-causalens (approved):
LGTM. What's the scenario where it would be updated to a string? Is it happening in user space? e.g. manual action update to a string rather than number

tamara-causalens (approved):

patricia-causalens:
> LGTM. What's the scenario where it would be updated to a string? Is it happening in user space? e.g. manual action update to a string rather than number

It was when it would get the value from the manager, it seems to store the values as string rather than number

claudiawestby:
> > LGTM. What's the scenario where it would be updated to a string? Is it happening in user space? e.g. manual action update to a string rather than number
> 
> It was when it would get the value from the manager, it seems to store the values as string rather than number, but if a user were to do it manually as a string I think it makes sense for the component to still display it

I don't believe the manager types it to a string (but if you find the place where it does this please let me know!). I believe this is due to the issue that I raised here: https://causalens.slack.com/archives/C04DH7ZKABZ/p1697455262967939

When I first saw this PR i thought it was to address this but am now realizing it is something different.

PR #134, opened 2023-10-24
Fix: DO-2027 none in numeric input shows 0

patricia-causalens:
## Motivation and Context

Claudia found that when starting NumericInput with Variable(), it displayed 0 instead of appearing empty

## Implementation Description

This was because Number(null) returns 0and not null. Although the variable contained the correct value the display did not match the same behaviour.

How the component behaves:
If starting with Variable() -> empty display
If Variable has a number -> displays the number
If Variable contains a string that can be converted to a number -> displays the number
If Variable contains something that cannot be converted to a number -> empty display

I checked that the above is followed for both initialising the Variable with those and by updating the Variable value through an action

None

## How Has This Been Tested?

Locally for all the scenarios covered above

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fixed an issue where `Input` of type number displayed zero istead of null when null value was set.
+
 ## 1.4.1
 
 -   Fixed issue where in some cases if `Input` of type number value variable was update outside the component, the value would not show.
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -4,7 +4,7 @@ title: Changelog
 
 ## NEXT
 
--   Fixed an issue where `Input` of type number displayed zero istead of null when null value was set.
+-   Fixed an issue where `Input` of type number displayed zero instead of null when null value was set.
 
 ## 1.4.1
 
--- a/packages/dara-components/js/common/input/input.tsx
+++ b/packages/dara-components/js/common/input/input.tsx
@@ -24,6 +24,9 @@ interface InputProps extends FormComponentProps {
 const StyledInput = injectCss(UiInput);
 const StyledNumericInput = injectCss(UiNumericInput);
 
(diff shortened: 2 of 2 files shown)


packages/dara-components/changelog.md
```diff
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fixed an issue where `Input` of type number displayed zero istead of null when null value was set.
```
krzysztof-causalens, line 7: typo `istead` -> `instead`

packages/dara-components/js/common/input/input.tsx
```diff
@@ -24,6 +24,9 @@ interface InputProps extends FormComponentProps {
 const StyledInput = injectCss(UiInput);
 const StyledNumericInput = injectCss(UiNumericInput);
 
+function getNumericValue(value: string | number): number {
+    return value == null || Number.isNaN(+value) ? null : +value;
```
krzysztof-causalens, line 28: What's the purpose of using `+` rather than `Number()` here?

krzysztof-causalens (commented):
Looks good, quick question

packages/dara-components/js/common/input/input.tsx
patricia-causalens: They both behave the same way here, I learned about it the other day and had just been looking for an excuse to try it.

packages/dara-components/changelog.md
patricia-causalens: oops fixed

claudiawestby (approved):
thanks so much :)

packages/dara-components/js/common/input/input.tsx
krzysztof-causalens: I would argue `Number` is more explicit, with `+` for example I had to think twice whether there wouldn't be some edge case with negative numbers.

patricia-causalens: That is fair, I have switched to use `Number`

krzysztof-causalens (approved):

PR #137, opened 2023-10-25
Improvement: DO-2030 normalize `py_component` responses

krzysztof-causalens:
## Motivation and Context

Dara implements a normalization mechanism, previously used for template response, and requests for DerivedVariables and py_components. This looks for instances of variables, replaces them with references and puts the actual values in a lookup. This decreases payload size in cases where the same variables are referenced in multiple places in a payload, especially when `Variable` has a large `default`.

This was previously not implemented for `py_component` results. A recent discovery in one of the apps was that a response for a particular `py_component` was about 100MB. This was due to the scenario described above, where a `Variable` default was large and duplicated dozens of times, as the component returned a list of cards, each having an action updating that variable. While this to some extent can be attributed to bad architecture/coding, this should still be handled on Dara side.

This PR extends the normalization mechanism to handle results of `py_components`, both for non-tasks and tasks.

## Implementation Description

`_make_render_safe` which is the final part of rendering a `py_component` has been updated to return a `NormalizedPayload`.

On the frontend we run the result of the component through `denormalize`.

## How Has This Been Tested?

Updated backend and frontend tests to use the new response structure.
Manually verified all types of py_component outputs work, including cases with tasks/progress trackers etc.

Verified in the app where the issue was found that this improvement reduces the request size from 100MB to 3.2MB.

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   `py_component` results are now normalized, which means multiple instances of the same variables within the returned components will be deduplicated. This should significantly reduce the amount of data sent over the wire in some cases.
+
 ## 1.4.2
 
 -   Fixed an issue where `.get()` API on `Variable` would not behave correctly when used as the update target within actions (either legacy `UpdateVariable` or new `ActionCtx.update` method)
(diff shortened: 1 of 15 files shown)


patricia-causalens (approved):
LGTM, nice work!

PR #147, opened 2023-10-31
Improvement: DO-2064 change plotly defaults

patricia-causalens:
## Motivation and Context

Some feedback from min-height for Plotly was that it was too small.
Another piece of feedback was that the default fonts were too large and that the plots had too large paddings by default

## Implementation Description

I have updated so that the default min-height is 350px. Have also added an example to docstrings about how this setting could be overwritten.

Based on feedback updated default `Plotly` settings to have reduced margins, and smaller fonts. Now by default title will take 16px which is our base font size, axis labels and legends will be one smaller at 14px.

So that they are consistent the same font changes were applied to Bokeh.

None

## How Has This Been Tested?

Locally on a test app

## PR Checklist:

Screenshot 2023-10-31 at 13 12 57


tamara-causalens (approved):
Lovely!

krzysztof-causalens (approved):

PR #149, opened 2023-11-01
Fix: DO-2052 radio group does not update when starting with an empty var

patricia-causalens:
## Motivation and Context

Noticed that when passing an empty Variable to a RadioGroup component in dara the Variable value was never updated.

## Implementation Description

When Variable was empty this line `items.find((item) => item.value === value)` returned `undefined`.
This meant the `RadioGroup` component behaved in uncontrolled mode and did not trigger on change events. Resulting in the variable never being updated.

## How Has This Been Tested?

Locally on a test app

## PR Checklist:


krzysztof-causalens:
> @krzysztof-causalens if someone were to pass a variable with some value that is not in the RadioGroup it would default to uncontrolled given the current implementation. Do you think that is ok or should we handle it differently?

I guess even if variable is not passed, `useVariable` falls back to a `useState` behind the scenes, so it should still be controlled. This way it will work within forms etc

krzysztof-causalens (approved):

PR #153, opened 2023-11-08
Improvement: DO-2127 Update --reload to infer correct directory, add --reload-dir flag

krzysztof-causalens:
## Motivation and Context

This PR brings improvement to the reloader configuration. Previously we were relying on uvicorn default watcher's behaviour defaults. This has proven to be problematic at times where it watches the entire CWD for `*.py` files, even outside of the source folder. 

The CLI now tries to infer the parent directory of the module containing the `config` variable which acts as the entrypoint. If the inference process fails for any reason, a warning is logged and the CLI falls back to the previous default.

In addition, the CLI also has a `--reload-dir` flag which can be used to override the default behaviour.

## Implementation Description

The inference utilises `find_spec` from `importlib` in order to find path to a module without executing it.

## How Has This Been Tested?

Tested locally on a test app, with or without an `__init__.py` (which affects some importlib behaviour). 

Screenshot 2023-11-08 at 09 30 35
Screenshot 2023-11-08 at 09 30 16

## PR Checklist:

### DIFF
--- a/docs/getting-started/local-setup.mdx
+++ b/docs/getting-started/local-setup.mdx
@@ -348,6 +348,28 @@ You can read more about virtual environments [on the `venv` package documentatio
 
 :::tip
 The `--reload` flag here will cause it to automatically reload the backend and UI whenever any changes are made and saved.
+
+By default it will infer and watch the parent directory of the module containing your `config` variable.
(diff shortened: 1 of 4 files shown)


tamara-causalens (approved):

patricia-causalens (approved):
LGTM

PR #156, opened 2023-11-09
Improvement: DO-2012 improve custom action impl handling

krzysztof-causalens:
## Motivation and Context

This PR brings small improvements to advanced use cases of the action API:

1. Added an `execute_action` method. This can be passed an arbitrary `ActionImpl` instance. The two main use cases are:
- composing external APIs - if an external function provides an `ActionImpl`, users can use this API to invoke it within their custom action
```python
def external_api(*args) -> ActionImpl:
  # something
  return NavigateTo('/some-url')


# user action
@action
async def my_action(ctx: actionCtx):
  # some custom logic
  await ctx.execute_action(external_api(*args))
```
- invoking custom impls

2. Improved `useAction` hook to invoke an optional `onUnhandledAction` callback for each impl encountered without a registered implementation. This combined with the API above allows users to pass arbitrary instructions down to custom components and handle them.

## Implementation Description

In the initial implementation `onUnhandledAction` was invoked once when an unknown action impl was encountered; this meant that potentially a stream of more instructions would stop and no longer be handled.

This has been moved deeper so each individual impl now falls back to using the `onUnhandledAction` callback if there is no registered implementation. 

## How Has This Been Tested?

Added Python and JS test cases

## PR Checklist:


patricia-causalens:
Could we add a small section in the actions page letting users know this exists as well? :)

tamara-causalens (approved):

Roman-Kornev (approved):

patricia-causalens (approved):

PR #158, opened 2023-11-21
Fix: DO-2141 Image and Anchor components base url handling

krzysztof-causalens:
## Motivation and Context

It was found that the Image component does not handle base urls properly, and appends the base url prefix even to external asset URLs.

Similar issue was found with the `Anchor` tag.

## Implementation Description

The utility used by the Image component now explicitly checks for `/static/` prefix (which is also validated on the backend) and applies the prefixing logic only to those requests.

The anchor component now uses the react router's `Link` component rather than an `<a>` tag for relative links since the router already handles the basename correctly.

## How Has This Been Tested?

Locally in embedded and standard environments

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/utils.ts
+++ b/packages/dara-core/js/utils.ts
@@ -1,5 +1,14 @@
+/**
+ * Prepend window.dara.base_url to asset path if defined.
+ * This is required for static assets to be loaded correctly when the app is not hosted at the root of the domain.
+ *
+ * Since all static assets are server from `/static/`, this is only applied to asset paths that start with `/static/`.
+ * This way we're not affecting any other assets that are loaded from other domains.
+ *
+ * @param asset asset URL
+ */
 export const prependBaseUrl = (asset: string): string => {
-    if (window.dara?.base_url) {
+    if (window.dara?.base_url && asset.startsWith('/static/')) {
         const baseUrl = new URL(window.dara.base_url);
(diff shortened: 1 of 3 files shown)


tamara-causalens (approved):
Looks good! Thanks!

packages/dara-core/js/utils.ts
```diff
@@ -1,5 +1,14 @@
...
+ * This is required for static assets to be loaded correctly when the app is not hosted at the root of the domain.
+ *
+ * Since all static assets are server from `/static/`, this is only applied to asset paths that start with `/static/`.
+ * This way we're not affecting any other assets that are loaded from other domains.
+ *
+ * @param asset asset URL
+ */
 export const prependBaseUrl = (asset: string): string => {
-    if (window.dara?.base_url) {
+    if (window.dara?.base_url && asset.startsWith('/static/')) {
```
patricia-causalens, line 11: If they configure static to be set as something else in config.static_files_dir this still results in being in static folder in here so that is fine, right?

patricia-causalens (approved):
LGTM

packages/dara-core/js/utils.ts
krzysztof-causalens: Yes, so `config.static_files_dir` changes which folder is mounted (default `dist`), but our backend router always mounts it at `/static` URL.

PR #159, opened 2023-12-04
Improvement: DO-2212 introduce request extras context

krzysztof-causalens:
## Motivation and Context

In certain situations it's desirable to uniquely identify where from on the page a Dara component/variable/action was called; this is not currently possible.

This PR implements a solution to that, where it is possible to inject `RequestExtras` into all requests made by Dara by simply wrapping parts of the tree in contexts:

```tsx
                <>
                    <RequestExtrasProvider
                        options={{
                            headers: {
                                'X-Dara-Test': 'test',
                            },
                        }}
                    >
                        <DaraTree />
                    </RequestExtrasProvider>
                    <RequestExtrasProvider
                        options={{
                            headers: {
                                'X-Dara-Test': 'test2',
                            },
                        }}
                    >
                        <DaraTree />
                    </RequestExtrasProvider>
                </>
```

This is not a breaking change, but obviously code relying on our implementation details might break. This should likely land in a minor release as it's a pretty significant internal refactor.

### `request` util changes

All requests in Dara are already made through the central `request` utility function, which is a light wrapper around `fetch`. This PR updates its last argument, which used to be a string (token) to also accept a `RequestOptions` object (in addition to string, for backwards compatibility). The object can contain both the token and extra options to be merged into the original options.

Changes have been made throughout Dara internals to update most places to use the new `RequestExtras = string | RequestOptions` type rather than a `token: string`. A new hook was added, `useRequestExtras` which pulls in session token and extras from nearest context - most usages of `useSessionToken` have been replaced with the above. 
Note this was not done in ALL places as sometimes we only cared about the token.

### variable hook changes

As the above was updated, an issue was discovered with the way the current `useVariable` hook and its internals are architected. Whenever a variable (or py_component, it works similarly) is first discovered, a new atom/selector is created for it and cached in a Map. The selectors though are essentially closures, so they capture the arguments passed into them when they were created. This means that the selectors capture `sessionToken` and requestExtras used **in the first place in the tree that they were passed into useVariable**. 
This was not an issue before as the token rarely changed. With the aforementioned changes though this meant that if the same variable was re-used in another part of the tree, only one set of request extras would always be used.

In order to be able to capture extras depending on the 'call site' of the variable, the hooks had to be modified.

#### DerivedVariable and `py_component internals`

These two were previously simple selectors. Because of the stale closure issue, they now utilize the [`selectorFamily` API](https://recoiljs.org/docs/api-reference/utils/selectorFamily). This means that rather than a static selector, they are now a function of `(requestExtras) => selector()`. The `useVariable` hook (and its internal methods like `getOrRegisterDerivedVariable`) now only register a *family* once per variable, and then call the variable with the fresh set of extras. Recoil then internally caches selector instances based on requestExtras passed.

Recoil compares arguments by value, so a `RequestExtrasSerializable` wrapper was created to adhere to Recoil's serializability needs (implements a `toJSON` method). 

Store internals were changed to keep track of `key => selectorFamily` and also `selectorFamily => Map(extras => selector)`. 

#### Variable

Variable used to be a simple `atom`, however it can relate to DerivedVariables through the `default` (i.e. `Variable.create_from_derived` API on the Python side). This means that if a DerivedVariable is used as a Variable's default, we need to again deal with the issue of using the correct selector from the DV's selectorFamily to use the correct requestExtras for the request.

The change was to again use an `atomFamily` instead in those cases, and instantiate it with the up-to-date extras. This way the default changed from

```tsx
default: getOrRegisterDerivedVariable(variable.default, ...)
```

to

```tsx
default: (requestExtras) => getOrRegisterDerivedVariable(variable.default, ..., requestExtras)
```

Note that because of the usage of `atomFamily`, each created instance of it will maintain its own pieces of state. This is obviously undesirable as we want variable state to be consistent on a page, regardless of the request params.
This is solved by using a simple effect attach to the atomFamily members. We're utilizing a simple `StateSynchronizer` singleton where we subscribe to notifications from other family members in order to propagate changes to one atom in a family to other family members. This behaviour has been covered by tests.

### other

I noticed the `useLocation` hook was used throughout the variable internals but was never actually used. Linters did not pick it up because of the circular dependencies. These are now removed which should help with performance especially when tweaking query params.

## How Has This Been Tested?

Added a few test cases covering the new behaviour of RequestExtras, as well as missing tests for the existing behaviour of `create_from_derived`.

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Internal (JS): added `RequestExtras` context to allow injecting additional e.g. headers into requests made by Dara in different parts of the component tree
+
 ## 1.4.6
 
 -   Added `execute_action` convenience function to `ActionCtx` to allow for executing an arbitrary `ActionImpl` instance. This can be useful in certain situations when an external API returns an action impl object to be executed, or for custom action implementations.
@@ -28,7 +32,6 @@ dara start --reload-dir decision_app/pages decision_app/utils
 
 -   Fixed an issue where the client app would refresh on WebSocket error&reconnect even without the `--reload` flag enabled
 -   Moved WebSocket server-side errors to be displayed on the default dev logger rather than the opt-in `--debug` logger
-
 
 ## 1.4.5
 
--- a/packages/dara-core/js/actions/download-variable.tsx
+++ b/packages/dara-core/js/actions/download-variable.tsx
@@ -124,10 +124,10 @@ const createMatrixFromValue = (val: Array<Record<string, any>> | any[][]): any[]
 const DownloadVariable: ActionHandler<DownloadVariableImpl> = async (ctx, actionImpl): Promise<void> => {
     let value = getVariableValue(actionImpl.variable, true, {
         client: ctx.wsClient,
+        extras: ctx.extras,
         search: ctx.location.search,
         snapshot: ctx.snapshot,
         taskContext: ctx.taskCtx,
-        token: ctx.sessionToken,
     });
 
     if (value instanceof Promise) {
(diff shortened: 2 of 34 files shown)


sam-causalens (approved):
LGTM, really nice job

patricia-causalens (approved):
LGTM!

Roman-Kornev (approved):

PR #162, opened 2023-12-05
Docs: DO-2193 pilot feedback on dara docs

patricia-causalens:
## Motivation and Context

Addressing some docs feedback received.

## Implementation Description

One of the feedback points included docs for the config. We should eventually prepend all the ones we don't intend people changing with underscore. For this however I have just added docs to the ones we expect users to change.

There were some broken links addressed.

It was said that would be nice to have a font size example on Text component.

And added a tip to use sizing_mode when using Bokeh as their plotting tool.

## PR Checklist:

### DIFF
--- a/docs/packages/dara-components/plotting/introduction.mdx
+++ b/docs/packages/dara-components/plotting/introduction.mdx
@@ -73,7 +73,7 @@ This comes from a requirement from Bokeh library itself, apps that use Bokeh fol
 Every time the recipe is used it must create a brand new collection of objects, this is so that there is no possibility for the recipe instructions to happen out of order.
 
 :::tip
-A useful prop to set to your `Bokeh` figure is `sizing_mode`, this allows for the `Bokeh` plot to adapt its size based on the space available to it.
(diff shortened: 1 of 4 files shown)


docs/packages/dara-components/plotting/introduction.mdx
```diff
@@ -17,7 +17,7 @@ In this page we will cover how to add the extension to your project and a quick
 
 ## Getting Started
 
-### Add your first `dara.components.plotting.bokeh.Bokeh` Component
+### Add your first `dara.components.plotting.bokeh.bokeh.Bokeh` Component
```
tamara-causalens, line 20: Does class linking work in the title? My experience was that it does not and that docusaurus gets a bit confused with `bokeh.bokeh` when it is in the title :( you can try to move the class linking bit in the text and leave the title as just text if it does not work :)

tamara-causalens (approved):

krzysztof-causalens (approved):

PR #165, opened 2023-12-06
Fix: DO-2171 cleanup of local storage

patricia-causalens:
## Motivation and Context

It was found that at some point local storage was running out of space. Seems like we were not cleaning up session related values.

## Implementation Description

Added so that we clean session related objects from Local storage.

## How Has This Been Tested?

Locally by using `Variable`s with persist flag and checking that storage was cleaned after log out/log in

Also updated a test for cleaning cache. Before it falsely succeeded due to waitFor not being properly awaited.

## PR Checklist:

### Before
not clean

### After
clear storage

### DIFF
--- a/packages/dara-core/js/shared/utils/clean-session-cache.tsx
+++ b/packages/dara-core/js/shared/utils/clean-session-cache.tsx
@@ -6,11 +6,19 @@
  */
 function cleanSessionCache(sessionToken: string): void {
     const sessionKeys = Object.keys(sessionStorage);
+    const localKeys = Object.keys(localStorage);
 
     sessionKeys.forEach((key) => {
         // Remove keys related to a different Dara session
         if (key.startsWith('dara-session') && !key.startsWith(`dara-session-${sessionToken}`)) {
             sessionStorage.removeItem(key);
+        }
+    });
+
+    localKeys.forEach((key) => {
+        // Remove keys related to a different Dara session such as persisted Variables
+        if (key.startsWith('dara-session') && !key.startsWith(`dara-session-${sessionToken}`)) {
+            localStorage.removeItem(key);
         }
     });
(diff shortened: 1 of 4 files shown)


packages/dara-core/js/shared/utils/clean-session-cache.tsx
```diff
@@ -6,13 +6,21 @@
  */
 function cleanSessionCache(sessionToken: string): void {
     const sessionKeys = Object.keys(sessionStorage);
+    const localKeys = Object.keys(localStorage);
```
krzysztof-causalens, line 9: Minor but this could be simplified if we just loop through the two storages, e.g.

```tsx
for (const storage of [sessionStorage, localStorage]) {
 const keys = Object.keys(storage);
 ...
}
```

packages/dara-core/tests/js/template-root.spec.tsx
```diff
@@ -31,21 +31,21 @@ describe('TemplateRoot', () => {
         expect(getByText('Frame, Menu')).toBeInstanceOf(HTMLSpanElement);
     });
 
-    it('should clean up cache on startup', () => {
+    it('should clean up cache on startup', async () => {
```
krzysztof-causalens, line 34: This test doesn't cover `sessionStorage` cleanup, could we have it cover both storages (or add another one to cover it)

krzysztof-causalens (changes requested):

packages/dara-core/tests/js/template-root.spec.tsx
patricia-causalens: oops had forgotten to add sessionStorage, has been added now

krzysztof-causalens (approved):

PR #167, opened 2023-12-07
Fix: DO-2259 action decorator not working in class and instance methods

### LINKED ISSUE #164: Defining @action in class layout
What is a recommended way of adding @actions in the class pages?

I have my variables and derived variables initialised in the `__init__` for full encapsulation but I am hitting a problem when trying to define an action

Have something like this:
```python
class Page:
    some_service: Service

    variable: Variable
    derived_variable: DerivedVariable

    def __init__(self, some_service: Service):
        self.some_service = some_service

        self.variable = Variable(default="default")

        self.derived_variable = DerivedVariable(
            func=self.some_service.get_derived_variable,
            variables=[self.variable],
        )

    @action
    async def reset_variable(self, ctx: action.Ctx, resolving_stage, selected_transaction):
        await ctx.update(variable=self.variable, value="")
 ```
 
 It doesnt like that there is a self in the function...
 
 Should it be something like this instead:
```python
variable = Variable(default="default")
  
@action
async def reset_variable(ctx: action.Ctx, resolving_stage, selected_transaction):
    await ctx.update(variable=self.variable, value="updated")

class Page:
    some_service: Service

    derived_variable: DerivedVariable

    def __init__(self, some_service: Service):
        self.some_service = some_service

        self.derived_variable = DerivedVariable(
            func=self.some_service.get_derived_variable,
            variables=[variable],
        )
 ```
 
 Its a bit annoying that I have to keep `Variable` and `DerivedVariable` defined separetly

krzysztof-causalens:
## Motivation and Context

Closes #164

Turns out the action decorator does not properly handle being used on bound methods as it does not currently take any extra measures to handle the `cls/self` parameter.

This PR updates the decorator to handle those cases, though at the expense of typing - expressing the possible parameter transformations including self/cls is not really possible in Python so the typing was made more lenient.

## Implementation Description

The decorator now checks the signature for whether the first param is `cls` or `self` to detect being used in bound methods. Unfortunately there isn't really a better way of doing that in Python, as the decorator is ran **before** a method is bound to the class/instance.

Added a `__get__` descriptor to the decorator as that's needed for Python to correctly inject the class or its instance.

The decorator `__call__` implementation was adjusted to store the bound argument in `static_kwargs` under a special prefixed key. The action execution logic then picks up that key and passes it as a first positional argument.

## How Has This Been Tested?

Extended test suite to cover cases where `@action` decorator is used as instance-, class- or static methods.

## PR Checklist:


patricia-causalens (approved):
LGTM

PR #168, opened 2023-12-08
Fix: DOS-479 Added support for multi websocket channels

jsumiguin-causalens:
## Motivation and Context

Currently, we treat web socket channels as a one-to-one relationship with the session. This works okay for most cases, but if we have two web socket channels for the same session(i.e. a different tab) values of the variables are lost as they are tied to the web socket channel.

## Implementation Description

- Added a `WS_CHANNEL` context var to websocket -- this is setup to get the actual variable on the actual channel
- Added functionality on `get_current_value` to get the current value on the channel if the channel is provided, otherwise, we get the value from all the channels.
- Changed the session_channel list to be a `Set` since one session can have multiple channels
- Trickled down session error to all channels

N/A
## How Has This Been Tested?

This has been tested locally and also by unit tests.
## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/interactivity/any_variable.py
+++ b/packages/dara-core/dara/core/interactivity/any_variable.py
@@ -135,12 +135,18 @@ async def get_current_value(variable: dict, timeout: float = 3, raw: bool = Fals
             return None
 
         session_channels: Dict[str, Set[str]] = {}
+        saved_ws_channel = WS_CHANNEL.get()
 
         # Collect sessions which are active
         for sid in session_ids:
             if websocket_registry.has(sid):
                 ws_channel = websocket_registry.get(sid)
-                session_channels[sid] = ws_channel
+                # If saved websocket channel is not none, then only save the session of that channel
+                if saved_ws_channel is not None:
+                    if saved_ws_channel in ws_channel:
+                        session_channels[sid] = {saved_ws_channel}
+                else:
+                    session_channels[sid] = ws_channel
 
         if len(session_channels) == 0:
             dev_logger.warning(
@@ -153,7 +159,6 @@ async def get_current_value(variable: dict, timeout: float = 3, raw: bool = Fals
 
         async def retrieve_value(channel: str):
             nonlocal registered_value_found
-
             try:
                 sentinel = object()
                 raw_result = sentinel
@@ -196,13 +201,8 @@ async def get_current_value(variable: dict, timeout: float = 3, raw: bool = Fals
 
         results = {}
 
-        saved_ws_channel = WS_CHANNEL.get()
-
         for session, channels in session_channels.items():
             for ws in channels:
-                # If we have a channel selected -- get just from that channel
-                if saved_ws_channel is not None and ws != saved_ws_channel:
-                    continue
 
                 raw_result = raw_results[ws]
                 # Skip values from clients where the variable is not registered
--- a/packages/dara-core/dara/core/interactivity/any_variable.py
+++ b/packages/dara-core/dara/core/interactivity/any_variable.py
@@ -134,7 +134,7 @@ async def get_current_value(variable: dict, timeout: float = 3, raw: bool = Fals
             )
             return None
 
-        session_channels: Dict[str, str] = {}
+        session_channels: Dict[str, set[str]] = {}
 
         # Collect sessions which are active
         for sid in session_ids:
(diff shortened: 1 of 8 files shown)


packages/dara-core/dara/core/interactivity/any_variable.py
```diff
@@ -134,7 +134,7 @@ def restore_contexts():
             )
             return None
 
-        session_channels: Dict[str, str] = {}
+        session_channels: Dict[str, set[str]] = {}
```
krzysztof-causalens, line 137: Please use uppercase `typing.Set` so this works with Python 3.8

packages/dara-core/dara/core/interactivity/any_variable.py
```diff
@@ -190,35 +190,41 @@ def is_valid(x):
...
+                # Skip values from clients where the variable is not registered
+                if raw_result == NOT_REGISTERED:
+                    continue
 
-            # Skip values from clients where the variable is not registered
-            if raw_result == NOT_REGISTERED:
-                continue
+                # If returning raw results, skip resolving at this point
+                if raw:
+                    results[session] = raw_result
```
krzysztof-causalens, line 208: Shouldn't this be `results[ws]` now?

packages/dara-core/dara/core/interactivity/any_variable.py
```diff
@@ -190,35 +190,41 @@ def is_valid(x):
                 raw_results[channel] = e
 
         async with anyio.create_task_group() as tg:
-            for chan in session_channels.values():
-                tg.start_soon(retrieve_value, chan)
+            for channels in session_channels.values():
+                for chan in channels:
+                    tg.start_soon(retrieve_value, chan)
```
krzysztof-causalens, line 195: If we have the channel selected, can we narrow this down to only send requests to that particular channel? And same in the loop processing the results below. We're ignoring other results anyway

We'll save a few requests and then we don't need the extra block below since we already return one value if only one result is present

packages/dara-core/tests/python/test_websocket.py
```diff
@@ -188,3 +194,229 @@ def handler(loop, context):
...
+
+            async with _async_ws_connect(testclient, token) as second_websocket:
+                second_init = await second_websocket.receive_json()
+                second_var_value = None
+                async def second_server():
+                    nonlocal second_var_value
+                    second_var_value = await variable.get_current_value()
+
+                ## Make sure the websocket channel is set correctly
+                WS_CHANNEL.set(second_init.get('message').get('channel'))
```
krzysztof-causalens, line 379: Can we have another test checking that if we don't set this variable, a dict is returned with both values?

krzysztof-causalens (changes requested):
Looks good Jon, thanks. Few minor improvements and one more test case to cover suggested below

packages/dara-core/dara/core/interactivity/any_variable.py
```diff
@@ -196,16 +196,22 @@ def is_valid(x):
 
         results = {}
 
+        saved_ws_channel = WS_CHANNEL.get()
+
         for session, channels in session_channels.items():
             for ws in channels:
+                # If we have a channel selected -- get just from that channel
+                if saved_ws_channel is not None and ws != saved_ws_channel:
```
krzysztof-causalens, line 204: This condition should also be applied in the loop above where WS messages are fired, so we only request variables from the channel we care about

jsumiguin-causalens: I've moved it now to the for loop that assigns the variable we're looping on. I think it doesn't affect anything else except for getting these values, but do let me know if there is any problem with having them there! Cheers.

packages/dara-core/dara/core/interactivity/any_variable.py
```diff
@@ -134,13 +134,19 @@ def restore_contexts():
...
             return None
 
-        session_channels: Dict[str, str] = {}
+        session_channels: Dict[str, Set[str]] = {}
+        saved_ws_channel = WS_CHANNEL.get()
 
         # Collect sessions which are active
         for sid in session_ids:
             if websocket_registry.has(sid):
                 ws_channel = websocket_registry.get(sid)
```
krzysztof-causalens, line 143: Can we rename this to `ws_channels`, it's a set so gets a little confusing to read

krzysztof-causalens (approved):
Looks good, last minor nitpick but otherwise good to go from my end.

patricia-causalens (approved):
LGTM

PR #176, opened 2024-01-18
Fix: DOS-577 Address failure of get_current_value

sam-causalens:
## Motivation and Context
Resolves an issue that caused variables to break when fetching their current value after the user navigates away and back to the application.

## Implementation Description
* Address the websocket session not being configured correctly in get_current_value after navigating away and back to a running kernel
* The double if block ended up with a situation where the sessions could be empty even though a session existed which lead to the request failing even though there was a connection to use.

No

## How Has This Been Tested?
Tested Locally

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-  Fixed an issue where the websocket channel would fail to be set correctly when using get_current_value after the user has reconnected their browser on a different websocket channel.
+
 ## 1.5.1
 
 -   Fixed an issue where Local storage was not being cleared between sessions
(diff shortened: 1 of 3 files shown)


krzysztof-causalens (commented):
Looks good, nice find. I believe we have some tests covering multi-user behaviour of this get_current_value/websockets piece (`tests/python/test_websocket.py`), could be worth adding a regression test case for this bug with same user reconnecting?

sam-causalens:
> Looks good, nice find. I believe we have some tests covering multi-user behaviour of this get_current_value/websockets piece (`tests/python/test_websocket.py`), could be worth adding a regression test case for this bug with same user reconnecting?

Yep, makes sense. Wasn't aware those tests were there.

krzysztof-causalens (approved):
> Yep, makes sense. Wasn't aware those tests were there.

Yeah the feature wasn't well tested initially (my fault) but @jsumiguin-causalens added a couple when extending it to support multiple channels per user. Thanks for adding one.

jsumiguin-causalens (approved):
Apologies for not spotting this sooner, good catch, though!

patricia-causalens (approved):
LGTM

PR #177, opened 2024-01-18
Feat: DO-2245: support time series in causal graph viewer

patricia-causalens:
## Motivation and Context

Updating Dara side to support tier changes made on dara-ui

## Implementation Description

Updated so that tiers can be passed through to the ui and update dara-ui dependency.
Added docs to cover usage of tiers in CausalGraphViewer and TimeSeriesCausalGraph

Also updated headings in graph introduction as up to h3 it still shows on sidebar, this will make the whole content of that page more discoverable.

## How Has This Been Tested?

Locally on a test app

## PR Checklist:


docs/packages/dara-components/graphs/introduction.mdx
```diff
@@ -336,3 +402,29 @@ config.add_page(name='Skeleton Example', content=CausalGraphViewer(causal_graph=
 ```
 
 ![CausalGraphViewer Skeleton Example](./assets/CausalGraphViewerSkeleton.png)
+
+### TimeSeriesCausalGraph
+
+For [TimeSeriesCausalGraph](https://causalgraph.causalens.com/docs/generated/cai-causal-graph/docs/time_series_causal_graph) we have an added in support where for layouts that support tiers, they will use `time_lag` and `variable_name` to place these nodes within the same tier.
```
krzysztof-causalens, line 408: Might be worth explaining the behaviour more, that e.g. passing explicit tier config will override the behaviour and this is just a default if a time series graph is detected and the layout supports it.

packages/dara-components/dara/components/graphs/graph_layout.py
```diff
@@ -57,6 +57,78 @@ class GraphLayout(BaseModel, abc.ABC):
     node_font_size: Optional[int] = None
 
 
+class TiersConfg(BaseModel):
+    """
+    TiersConfg provides a way of defining tiers for a graph layout.
+
+    :param group: Path within node to group property which defines the tier it belong to
+    :param order_nodes_by: A path to a node property which contains a number defining the order of nodes within a tier
+    :param rank: A list of grouo values defining the order they should appear in
```
krzysztof-causalens, line 66: Typo `grouo` -> `group`

krzysztof-causalens, line 66: Might be worth adding a very short example in the docstring so users can understand what these mean at a glance, I know you have a long one in TieringLayout but users likely won't see the docstring in their editor as this is just a base class. TiersConfig is a class they will interact with so could be useful to add

packages/dara-components/js/graphs/graph-layout.tsx
```diff
@@ -138,6 +142,18 @@ export function parseLayoutDefinition(definition: GraphLayoutDefinition): GraphL
...
+            }
+
+            if (definition.tiers) {
+                builder.tiers = definition.tiers;
+            }
+
+            if (definition.orientation) {
+                builder.orientation = definition.orientation;
+            }
+
```
krzysztof-causalens, line 156: Could we extract the repeated tier properties somehow? e.g. something like

```tsx
    let builder;
    switch (definition.layout_type) {
        case '{layout_type}': {
            builder = LayoutClass.Builder
            // ...
            break;
        }
        // ...
   }

   if (<builder class supports tiers>) {
     // assign tier properties
   }
```

krzysztof-causalens (approved):
Looks good, few minor comments but good to go otherwise

tamara-causalens (approved):
Very nice! Agree with the docs comments, looks good to me after that is addressed 😊

packages/dara-components/js/graphs/graph-layout.tsx
patricia-causalens: I have attempted this, could you double check? Planar works a bit different so I left that separate

packages/dara-components/js/graphs/graph-layout.tsx
```diff
@@ -9,10 +9,18 @@ import {
...
 // Types mirror backend types defined in dara.components.graphs.graph_layout
 type GraphLayoutType = 'marketing' | 'planar' | 'spring' | 'circular' | 'fcose' | 'force_atlas' | 'custom';
 
+interface DefinitionWithTiers extends TieredGraphLayoutBuilder {
+    tier_separation?: number;
+}
+
+interface BuilderWithTiers extends TieredGraphLayoutBuilder {
+    tierSeparation?: (separation: number) => GraphLayoutBuilder<unknown>;
+}
```
krzysztof-causalens, line 23: This could be defined next to the builders and extended directly by the tier-supporting builders

packages/dara-components/js/graphs/graph-layout.tsx
```diff
@@ -258,6 +231,21 @@ export function parseLayoutDefinition(definition: GraphLayoutDefinition): GraphL
         }
     }
 
+    if (isDefinitionWithTiers(definition)) {
+        const builderWithTiers = builder as unknown as BuilderWithTiers;
```
krzysztof-causalens, line 235: You could force TypeScript to treat `builder` as `BuilderWithTiers` by building another typeguard, so you'd do `isDefinitionWithTiers(definition) && isTiersBuilder(builder)`, if the BuilderWithTiers was recognizable from its shape (if not can attach a 'supportsTiers' property or similar)

krzysztof-causalens (commented):
I think this is fine for now, could use some cleanup in the future though as the type names and placement is confusing in places

packages/dara-components/js/graphs/graph-layout.tsx
```diff
@@ -9,10 +9,18 @@ import {
     SpringLayout,
     TieredGraphLayoutBuilder,
 } from '@darajs/ui-causal-graph-editor';
+import { GraphLayoutBuilder } from '@darajs/ui-causal-graph-editor/dist/shared/graph-layout/common';
 
 // Types mirror backend types defined in dara.components.graphs.graph_layout
 type GraphLayoutType = 'marketing' | 'planar' | 'spring' | 'circular' | 'fcose' | 'force_atlas' | 'custom';
 
+interface DefinitionWithTiers extends TieredGraphLayoutBuilder {
```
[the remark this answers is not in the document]
patricia-causalens: Makes sense, will add a cleanup ticket on the ui side with these feedbacks and then return to this once that is addressed

PR #179, opened 2024-01-18
Docs: DO-2316: updating layout docs

patricia-causalens:
## Motivation and Context

Some people had difficulties with hug and Graph/Table components. Based on feedback it was worth adding a small section in the docs briefly explaining the issue.

## Implementation Description

Updated the docs on lay it out page to include explanation of why some components became so small within a hug component.

## PR Checklist:

### DIFF
--- a/docs/best-practices/laying-out-apps.md
+++ b/docs/best-practices/laying-out-apps.md
@@ -163,7 +163,7 @@ Stack(
 
 :::caution
 
-**When dealing with `CausalGraphViewer`, `Table` and `Scrollable Stacks`:**
+**When dealing with `CausalGraphViewer`, `Table`, `Plotly` and `Scrollable Stacks`:**
 These components are special in that they essentially "scan" the space available on the page before they render. This is good because it allows them to adapt to their surroundings but it comes with drawbacks.
 
 If you surround them with hug, hug tries to make the space only as big as it needs to. So there is a conflict of interest as there is no space leftover for them to take. In that case they will take their default `min_height`. So if you are using these components within a `hug` container you may need to overwrite `min_height` of the component should you want them to take more space on the screen.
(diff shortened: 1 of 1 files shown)


tamara-causalens (approved):
Great!

krzysztof-causalens (approved):

docs/best-practices/laying-out-apps.md
```diff
@@ -161,6 +161,14 @@ Stack(
 
 ```
 
+:::caution
+
+**When dealing with `CausalGraphViewer`, `Table`, `Plotly` and `Scrollable Stacks`:**
+These components are special in that they essentially "scan" the space available on the page before they render. This is good because it allows them to adapt to their surroundings but it comes with drawbacks.
+
+If you surround them with hug, hug tries to make the space only as big as it needs to. So there is a conflict of interest as there is no space leftover for them to take. In that case they will take their default `min_height`. So if you are using these components within a `hug` container you may need to overwrite `min_height` of the component should you want them to take more space on the screen.
```
claudiawestby, line 169: "If you surround them with hug" this sounds a bit out of context. Can we be more explicit ie "If you place them in a container that has `hug` set to `True`"

docs/best-practices/laying-out-apps.md
```diff
@@ -161,6 +161,14 @@ Stack(
 
 ```
 
+:::caution
+
+**When dealing with `CausalGraphViewer`, `Table`, `Plotly` and `Scrollable Stacks`:**
```
claudiawestby, line 166: If we put scrollable stacks in capital letters and within `code brackets` it seems like it is its own component which could send someone looking for it. could we do scrollable `Stack`s?

docs/best-practices/laying-out-apps.md
patricia-causalens: Good point, addressed

docs/best-practices/laying-out-apps.md
patricia-causalens: fair point, fixed

PR #181, opened 2024-01-23
Fix: DO-2346 select with items initial value displays wrong label

patricia-causalens:
## Motivation and Context

An issue was found with the `Select` component where if the items in it were set as a list of `Item`s, and the initial value was just a number value, then it displayed the number value in the input rather than the label.

## Implementation Description

The issue was found in the `toItem` method which got the value and created a new `Item` object instead of getting this from the defined/derived items array. This meant that if label and value did not match then the wrong label was shown to the user

None

## How Has This Been Tested?

Locally in the scenario in which the issue was initially found, by varying initial value of being an `Item`, a value and differently defined items passed.

## PR Checklist:


packages/dara-components/js/common/select/select.tsx
```diff
@@ -38,9 +38,6 @@ export function getMultiselectItems(values: Array<any>, items: Array<Item>): Arr
     }, []);
 }
 
-const toItem = (value: any): Item | undefined | null =>
-    typeof value === 'string' || typeof value === 'number' ? { label: String(value), value } : value;
-
```
krzysztof-causalens, line 43: I think it was intended that this did not look up existing items here. The use case was for persisting selected value when the item was no longer (perhaps temporarily) in the list. I'd speak to @Roman-Kornev about it as I think he was implementing it for our internal product here https://github.com/causalens/dara/pull/123, let's make sure it still serves that purpose

packages/dara-components/js/common/select/select.tsx
```diff
@@ -53,6 +50,11 @@ function hasListSection(items: Item[] | ListSection[]): items is ListSection[] {
 function isStringArray(value: any): value is string[] {
     return Array.isArray(value) && value.every((item) => typeof item === 'string');
 }
+
+/** Type guard to check if an object is an Item */
+function isItem(obj: any): obj is Item {
+    return Object.prototype.hasOwnProperty.call(obj, 'value');
```
krzysztof-causalens, line 56: Could also add `typeof obj === 'object'`, otherwise might call `value` on null and throw an error

packages/dara-components/js/common/select/select.tsx
patricia-causalens: Oh I see, will update so that if the value is not present it can still be shown

packages/dara-components/js/common/select/select.tsx
patricia-causalens: good point, fixed

packages/dara-components/js/common/select/select.tsx
```diff
@@ -95,6 +97,22 @@ function Select(props: SelectProps): JSX.Element {
...
+     */
+    const toItem = (val: any): Item | undefined | null => {
+        // We set initially the item as an Item type with the value as the label, so that select can show a value even if not present in the list
+        let foundItem = typeof val === 'string' || typeof val === 'number' ? { label: String(val), val } : val;
+        // Tries to get matching item based on value from formattedItems
+        const matchingItem = formattedItems.find((item) => isItem(item) && String(item.value) === String(val));
+        // If a matching Item is found, update foundItem
+        if (matchingItem) {
+            foundItem = matchingItem;
+        }
```
Roman-Kornev, line 112: If this always takes priority I suggest putting this at the top of the function and early retrning. Then we can just return `foundItem` if it didn't match.

Roman-Kornev (approved):

packages/dara-components/js/common/select/select.tsx
Roman-Kornev: Yes, that's correct. It was for preserving values if they are not present in the list

packages/dara-components/js/common/select/select.tsx
krzysztof-causalens: Actually, sorry I blanked on it when commenting - `typeof null` resolves to `object` as well (because JavaScript) so you might need to explicitly check for null

packages/dara-components/js/common/select/select.tsx
patricia-causalens: fair point, have updated this

packages/dara-components/js/common/select/select.tsx
```diff
@@ -95,6 +97,21 @@ function Select(props: SelectProps): JSX.Element {
...
     }, [items]);
 
+    /**
+     * Converts a value to an Item, or finds a matching Item in formattedItems based on value.
+     * @param val - The value to be converted or matched.
+     */
+    const toItem = (val: any): Item | undefined | null => {
+        // Tries to get matching item based on value from formattedItems
+        const matchingItem = formattedItems.find((item) => isItem(item) && String(item.value) === String(val));
+        // If a matching Item is found, update foundItem
```
krzysztof-causalens, line 107: The comment seems to be out of date?

packages/dara-components/js/common/select/select.tsx
```diff
@@ -53,6 +50,11 @@ function hasListSection(items: Item[] | ListSection[]): items is ListSection[] {
 function isStringArray(value: any): value is string[] {
     return Array.isArray(value) && value.every((item) => typeof item === 'string');
 }
+
+/** Type guard to check if an object is an Item */
+function isItem(obj: any): obj is Item {
+    return obj && Object.prototype.hasOwnProperty.call(obj, 'value');
```
krzysztof-causalens, line 56: I've commented on this in the previous thread but it probably got lost in the chain:

`typeof null` resolves to `object` as well (because JavaScript) so you might need to explicitly check that `obj` is not null as well

patricia-causalens: but if obj is null won't the first part of the condition fail?

packages/dara-components/js/common/select/select.tsx
patricia-causalens: Oops updated

packages/dara-components/js/common/select/select.tsx
krzysztof-causalens: Right sorry I missed your last update 🤦 ignore my comment, this is fine now :)

PR #182, opened 2024-01-23
Improvement: DO-2345 layout components other than stack not accepting justify align

patricia-causalens:
## Motivation and Context

Updating layout components so that they behave more in line with Stack. Often people had to add Stacks within these to apply basic align/justify css which really isn't necessary and just adds layers of unnecessary stack to their apps.

## Implementation Description

- Added `justify` and `align` to be accepted by `LayoutComponents` on Python side and updated docstring to all of those that support it  so that it is more accessible.
- Added a new interface for `LayoutComponentProps` on js side which does the same.
- Updated for `align` to accept a string, it is in line with other definitions.
- Removed the conversion for `align` on Stack, `flex-start` is different than `start` when `row-reverse` is set.  And if the html `dir` attribute is defined it can cause `start` to align differently to `flex-start`. Otherwise they behave the same, so the conversion that we had on `Stack` did not seem needed.
- Updated so that `Modal`, `Card`, `Grid` and `Form` will now all accept align and justify.
- For `Card` this is applied at the content level so that it does not affect the title/subtitle.
- For `Grid.Column` this involved updating one of the props names to be just align, more consistent with other components.

None

## How Has This Been Tested?

Locally on a test app

## PR Checklist:


packages/dara-components/dara/components/common/base_component.py
```diff
@@ -45,9 +45,15 @@ class LayoutComponent(BaseDashboardComponent):
...
     append/pop functionality that allows for these components to be created more dynamically (e.g. inside a loop)
+
+    :param position: the position of the component, defaults to 'relative'
+    :param justify: the justify-content value to be passed to the component
+    :param align: the align-items value to be passed to the component
     """
 
     position: str = 'relative'
+    justify: Optional[str] = None
+    align: Optional[str] = None
```
krzysztof-causalens, line 56: We could make those literals for better typing, e.g. can copy the types from `dara.core.visual.css.CSSProperties`

packages/dara-core/js/types/core.tsx
```diff
@@ -188,7 +188,7 @@ export interface BaseComponentProps {
  */
 export interface BaseStylingProps {
     /** How to align component, either itself or children within */
-    align?: 'start' | 'end' | 'left' | 'right' | 'center' | 'justify' | 'match-parent';
+    align?: string;
```
krzysztof-causalens, line 191: Isn't `align` also a layout component specific prop? Why does it belong to the base styling props? Doesn't seem to be handled by `useComponentStyles` so should probably be moved as well

patricia-causalens: Other components may also have `align` for example you can apply it to `Text` which is not a layout component

packages/dara-components/dara/components/common/base_component.py
patricia-causalens: Nice! Forgot we had those :)

packages/dara-components/dara/components/common/base_component.py
```diff
@@ -15,10 +15,50 @@
 limitations under the License.
 """
 
-from typing import Optional
+from typing import Optional, Literal
 
 from dara.core.definitions import ComponentInstance, StyledComponentInstance
 
+JustifyTypes = Literal[
```
krzysztof-causalens, line 22: Same concepts as with enums, types should also be singular unless they represent a list of some sort, here's it a single value (but a union of possible options). So `JustifyType` and `AlignType`, or I could see an argument to just name it after the property name e.g. `JustifyContent` / `AlignItems`

packages/dara-core/js/types/core.tsx
krzysztof-causalens: Shouldn't Text define align then, as it's a different align (`text-align`) than the flex align (`align-items`) I believe?

patricia-causalens:
On the text component it does result in text-align. I didn't remove it from here because other components can define align differently and it was not unique to layout. But we can have it individually defined if not in layout components as well.

packages/dara-core/js/types/core.tsx
patricia-causalens: On the text component it does result in text-align. I didn't remove it from here because other components can define align differently and it was not unique to layout. But we can have it individually defined if not in layout components as well.

krzysztof-causalens: Right. Perhaps that's a future improvement, we should eventually disentangle these, they really shouldn't have the same name.

patricia-causalens: Yes we can think about it and see what we want to do :) 
I do see an argument of it being called the same for simplicity since it is the same concept and one thing to understand and remember for someone not familiar with css.  But also adds a bit of magic of it being translated to different things underneath depending on the component it is used on.

packages/dara-components/dara/components/common/base_component.py
patricia-causalens: Oh ok, thank you!

claudiawestby (approved):

packages/dara-core/js/types/core.tsx
krzysztof-causalens: Sure, but at the very least the prop should be defined on Text itself explaining what it maps to rather than on the very base components, as some components might not implement align at all.

PR #183, opened 2024-01-23
Improvement: DO-2402 add min height to bokeh component

patricia-causalens:
## Motivation and Context

When container surround `Bokeh` had `hug` set to `True` and figure had `sizing_mode` set to `scale_both` or similar then the plot would not be visible.

## Implementation Description

Updated so that `Bokeh` will have `min-height` and `min-width` set in line with `Plotly` in those scenarios.
By default if the user does not do anything with `Bokeh` it will have a fixed size on the page, as figure does.
If they set any other height within the figure that will still be respected, but if that value is smaller than the `min_height` of 350px the user would have some leftover space unless they updated this value on the `Bokeh` component itself and not the figure.

None

## How Has This Been Tested?

Locally on a test app

## PR Checklist:


tamara-causalens (commented):
Excellent!

PR #185, opened 2024-01-29
Improvement: DO-2342 inherit suspend_render setting

krzysztof-causalens:
## Motivation and Context

Current behaviour of `suspend_render` was to only apply it to a specific component. This resulted in not-so-great DX as it often needed to be repeated on children of a given component when the behaviour was desired for the entire sub-tree.

The behaviour is now changed in that the setting is now inherited and applied "recursively" to an entire sub-tree, i.e. all children, grandchildren and so on of a given component. This is technically a change in behaviour but since it's a rather advanced feature and the new behaviour is more desirable, we can release it in a minor/patch as it shouldn't be breaking.

## Implementation Description

This is implemented by hooking into the FallbackContext in DynamicComponent. The value provided to the context is now taken from the parent context value, if present and not overriden by the user.

## How Has This Been Tested?

Locally
Added unit test

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -5,6 +5,7 @@ title: Changelog
 # NEXT
 
 -   Fixed an issue where import discovery would consider the same symbols repeatedly causing it to run much longer than necessary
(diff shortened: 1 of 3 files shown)


Roman-Kornev (approved):
LGTM

patricia-causalens (approved):
LGTM

PR #186, opened 2024-01-29
Fix: DO-2458 Setting default theme for Plotly had wrong import

tamara-causalens:
## Motivation and Context
Plotly docs had a wrong import for `set_default_plotly_theme`. I've also added an example on how to use it

## PR Checklist:

### DIFF
--- a/docs/packages/dara-components/plotting/introduction.mdx
+++ b/docs/packages/dara-components/plotting/introduction.mdx
@@ -169,10 +169,27 @@ You should use:
 You can choose a default theme for all your plotting components using the following
 
 ```python
-from dara.components.plotting import set_default_bokeh_theme, set_plotly_default_theme
+from dara.components.plotting import set_default_bokeh_theme, set_default_plotly_theme
 
 # provide theme object as expected by bokeh Document, see docs https://docs.bokeh.org/en/2.4.1/docs/reference/document.html#bokeh.document.Document.theme
 set_default_bokeh_theme({ ... })
 
 set_default_plotly_theme({ ... })
 ```
+
+For example, for Plotly, see the following example:
+
+```python
+import plotly.express as px
+from dara.components import Plotly
+from dara.components.plotting import set_default_plotly_theme, dark_theme
+
+
+df = px.data.iris()  # Using built-in Iris dataset
+fig = px.scatter(df, x='sepal_width', y='sepal_length', color='species', title='Iris Dataset Scatter Plot')
+
+
+set_default_plotly_theme(theme=dark_theme)
+Plotly(figure=fig, min_height=500)
+```
+


krzysztof-causalens (approved):

docs/packages/dara-components/plotting/introduction.mdx
```diff
@@ -169,10 +169,27 @@ You should use:
...
 
 set_default_plotly_theme({ ... })
 ```
+
+For example, for Plotly, see the following example:
+
+```python
+import plotly.express as px
+from dara.components import Plotly
+from dara.components.plotting import set_default_plotly_theme, dark_theme
```
patricia-causalens, line 185: This is not quite the correct import because there exists a `dark_theme` for both `Plotly` and `Bokeh` and these are different. So I am not sure which one would get prioritised in the import and if this works, something to look into.

So if using plotly to guarantee the correct one should be:
`from dara.components.plotting.plotly import dark_theme`

We can add a note explaining where to get these themes from and that they could use this as a base to tweak and make their own

patricia-causalens: Also you can import `set_default_plotly_theme` from top level with the plotly import if you want

krzysztof-causalens: Good catch, as for which one is prioritised - the one which is present in `plotting/__init__.py` second would take priority (at the moment it's plotly). It's unfortunate both have the same name, I guess best thing to do is to use the more specific import as you've said

tamara-causalens: done!

docs/packages/dara-components/plotting/introduction.mdx
```diff
@@ -169,10 +169,27 @@ You should use:
...
-from dara.components.plotting import set_default_bokeh_theme, set_plotly_default_theme
+from dara.components.plotting import set_default_bokeh_theme, set_default_plotly_theme
 
 # provide theme object as expected by bokeh Document, see docs https://docs.bokeh.org/en/2.4.1/docs/reference/document.html#bokeh.document.Document.theme
 set_default_bokeh_theme({ ... })
 
 set_default_plotly_theme({ ... })
 ```
+
+For example, for Plotly, see the following example:
```
patricia-causalens, line 180: I think it is still worth adding a little sentence or two here saying what `set_default...` takes, which is just s dict representation of the theme that each plotting library takes. Since they are just dicts people can easily add to them should they wish. 

Some references that could be nice to add:
For Plotly: https://plotly.com/javascript/reference/#layout
For Bokeh: https://docs.bokeh.org/en/latest/docs/reference/themes.html#theme

patricia-causalens (approved):

PR #187, opened 2024-01-30
Fix: DO-2441 tiers confg has a typo on dara and add layering algo prop

patricia-causalens:
## Motivation and Context

Exposing layering_algorithm on dara side and fixinf a typo where TiersConfg was missing the i

## Implementation Description

Replaced TiersConfg name to TiersConfig
Updated ui deps to 1.5.2
Added layering_algorithm to the builder for PlanarLayout

dara-ui updated to 1.5.2

## How Has This Been Tested?

Locally on a demo app

## PR Checklist:


pnpm-lock.yaml
krzysztof-causalens, line 1: Why is the diff here so large? Just curious, can't view the diff on GitHub because of the size 😅

krzysztof-causalens (approved):

pnpm-lock.yaml
krzysztof-causalens: 🤔 you shouldn't be running the lockfile through the formatter, can you include the original generated by pnpm?

patricia-causalens: Sorry that was the originally created, I didn't run anything one it, it was created by the update-ui-deps command. I think the formatter is a separate issue I am having.

krzysztof-causalens: Right, I see. Okay let me have a look at the diff locally just as a sanity check nothing weird happened, it is unusually large

krzysztof-causalens: It seems like the previous lockfile version on master was autoformatted in #177 so this is just undoing it, this is fine :)

PR #189, opened 2024-02-06
Fix: DOS-625 Address action execution blocking new requests

sam-causalens:
## Motivation and Context
* An internal project was stalling requests when multiple actions were in flight that led to the app appearing broken for a time until the first action completed.

## Implementation Description
* The usage of BackgroundTasks in the new execute_action code was found to be causing blockages of new requests being picked up by the event loop. I think this maybe a reccurance of https://github.com/encode/starlette/issues/919 which they claim is resolved, but still doesn't appear to be working for us. It may be due to one of our middlewares behaving in a particular way that causes this, but the fix applied here appears to be the simplest fix for now.
* In this instance we have replaced the usage of background_tasks.add_task with asyncio.create_task that will run the task in the loop without worrying about its result.

No

## How Has This Been Tested?
Should be covered by existing unit tests

## PR Checklist:

Reproduction Code (in a dara app): 
```python
@action
async def test_action(ctx):
    print(f'running action {ctx}')
    for i in range(20):
        await anyio.sleep(0.2)
    print(f'finished action {ctx}')

config.add_page('action bug', Stack(*[Button(f'action-{x}', onclick=test_action(x)) for x in range(10)]))
```

Behaviour Before: 
https://github.com/causalens/dara/assets/58470795/8ac1fef4-4505-4a39-a911-30fe8f794b7d

Behaviour After: 
https://github.com/causalens/dara/assets/58470795/bd65f4b2-dc28-4669-8bbd-5b0827fe4f62

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-  Address action execution blocking new requests due to an issue around BackgroundTask processing in starlette
+
 ## 1.6.0
 
 -   Fixed an issue where import discovery would consider the same symbols repeatedly causing it to run much longer than necessary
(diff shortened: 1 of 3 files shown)


krzysztof-causalens (approved):
LGTM, thanks

PR #191, opened 2024-02-07
Fix: DOS-625 fix the test that became flaky on CI after recent change to action execution

sam-causalens:
## Motivation and Context
Address the flaky test when running in CI

## Implementation Description
Use wait_assert to wait for the call to be processed rather than expecting it to be called immediately.

No

## How Has This Been Tested?
Needs to be tested in CI as it passed everytime locally.

## PR Checklist:


krzysztof-causalens (approved):
LGTM, thanks for fixing it

tamara-causalens (approved):
Thanks!

PR #195, opened 2024-02-09
Fix: DOS-649 Properly recreate the websocket client when it is closed

sam-causalens:
## Motivation and Context
Found this issue whilst debugging studio being unresponsive after being left for a while on another tab.

## Implementation Description
* If the websocket client is closed for some external reason, e.g. browser going to sleep or a crash being thrown down it then the newly created websocket client is now properly reconnected to all the relevant streams and channels.

No

## How Has This Been Tested?
Tested Locally in a running studio instance

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/shared/template-root/template-root.tsx
+++ b/packages/dara-core/js/shared/template-root/template-root.tsx
@@ -86,7 +86,7 @@ function TemplateRoot(): JSX.Element {
                 setWsClient(setupWebsocket(token, config.live_reload));
             };
         }
-    }, [token, config.live_reload, wsClient]);
+    }, [token, config?.live_reload, wsClient]);
 
     if (templateLoading || actionsLoading || componentsLoading) {
         return null;
--- a/packages/dara-core/js/shared/template-root/template-root.tsx
+++ b/packages/dara-core/js/shared/template-root/template-root.tsx
@@ -78,16 +78,6 @@ function TemplateRoot(): JSX.Element {
         };
     }, [token, config?.live_reload]);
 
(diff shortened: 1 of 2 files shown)


packages/dara-core/js/shared/template-root/template-root.tsx
```diff
@@ -78,6 +78,16 @@ function TemplateRoot(): JSX.Element {
         };
     }, [token, config?.live_reload]);
 
+    // Effect that registers an onclose event handler on the newly created websocket each time its remade. It will then
+    // call setState to trigger an update of the client to all subscribing components so they have the new connection
```
krzysztof-causalens, line 82: Are you sure this is enough to make it work? A lot of internal functionality references the WsClient and does not really handle it being re-made, e.g. variables which only register atoms/selectors once. Perhaps it would be better if we could 'fix' the same ws client instance and update its internal socket rather than making another client?

sam-causalens: Makes sense, I wasn't aware of all the internals. I've updated the approach to reconnect everything within the same client instance. The only thing that will change is the channel, so I added an extra observable that can be used as a trigger, but from what I could see most things get the channel off the websocket client during execution so they should now get the updated value. 

I checked it in dStudio and actions and variables still seem to work correctly after the reconnect.

krzysztof-causalens: Makes sense. That's correct, most internals just interact with the methods to retrieve the channel/observables for certain message scopes so as long as the client reference stays the same, it should be fine. Thanks.

PR #197, opened 2024-02-12
Feat: DO-2501 collaboration stage 1 - add `Variable.store` prop, `BackendStore`, `InMemoryBackend`

krzysztof-causalens:
## Motivation and Context

This PR implements Stage 1 of the collaboration feature as per spec (see Jira ticket).

The collaboration feature is implemented in the form of a new `store` prop for the `Variable` class which accepts a `PersistenceStore` implementation. 

Note that this is an initial implementation with minimal set of features, so some of the details of the API are subject to change - this is marked in the docstring and changelog.

The initial implementation includes:
- `BackendStore` - `PersistenceStore` implementation for backend-based storage, accepts a `PersistenceBackend` implementation for storage
- `InMemoryBackend` - simple `PersistenceBackend` implementation

The user facing API is fairly simple:

```python
```python
from dara.core.persistence import BackendStore, InMemoryBackend
from dara.components import Input

# Store have a unique identifier and a backend implementation for storage
collab_variable = Variable(default=1, store=BackendStore(uid='my_variable', backend=InMemoryBackend()))

# Input value will automatically be kept in sync 'live' for all users
Input(value=collab_variable)


# The store can be interacted with directly
store: BackendStore = collab_variable.store
# Write a new value to the store, all clients will be notified of the change
await store.write('new value')
# Read the current value from the store
value = await store.read()
```

Note that due to typing limitations to properly infer the type of the store used we would have to introduce a second generic parameter for the `Variable` class. This would be a breaking change, as Python does not support defaults for generic arguments [before version `3.12`](https://peps.python.org/pep-0696/) (`mypy` also doesn't have a stable release supporting it as of the time of creating the PR either). The `default` kwarg in `TypeVar` is supported in `typing_extensions` but not passing the second generic fails at runtime regardless so it doesn't help.

In order to not introduce a breaking change I've opted not to introduce the generic parameter which results in worse typing, as the variable could technically have any PersistenceStore implementation (or `None`) attached:

Screenshot 2024-02-12 at 15 23 25

So the typing workaround is to use `cast` when interacting with a backendstore, if type-safety is desired:

```python
current_value = await cast(BackendStore, var.store).read()
```

Did not add a documentation page, will be added separately as we flesh out, test and stabilize the API via usage in internal product.

### Backend

When a `BackendStore` is instantiated, an entry is added to a new registry which enables the standard `RegistryLookup`-based customization for advanced use cases.

Invoking `write` or `delete` on `BackendStore` broadcasts the new value to all clients via the existing WS connection, as a new internal message type.

Two new endpoints have been added:
- `POST /api/core/store` - accepts a 'diff' - object of shape `{ 'store_uid': 'value' }` which enables syncing multiple store values in a single request, shape is determined by the Recoil API used which can sometimes choose to batch to the store; invokes the `write` method for each store
- `GET /api/core/store/{store_uid}` - retrieve the current value of the given store; invokes the `read` method on the backend attached to a given store

### Frontend

Syncing is implemented via the official `recoil-sync` library which we already use for URL variables. 
In this PR I implemented a new recoil store as per https://recoiljs.org/docs/recoil-sync/implement-store/ for `BackendStore`, as well as migrating the previous `persist_value` implementation to consolidate. This will make it easier to later consolidate the APIs further when we deprecate and eventually remove `persist_value` in favour of using a custom store.

The `useVariable` implementation for plain variable was updated to now always use an atom family rather than having two branches. This simplifies the logic, and was required as now the particular store implementation (like the `BackendStore`) might need to make requests to remote which needs to pull from latest `RequestExtras` provided in the given part of the tree. 

## How Has This Been Tested?

Locally (see video below)
Unit tests added 

## PR Checklist:

https://github.com/causalens/dara/assets/87647189/cb5bfca4-6996-4189-9ec5-50a461554ccc


packages/dara-core/js/api/http.tsx
```diff
@@ -48,6 +49,19 @@ export class RequestExtrasSerializable {
...
+        return serializable;
+    }
+
+    /**
+     * Serialize the extras into a string.
+     */
+    toJSON(): string | null {
+        const serializable = this.toSerializable();
+
+        if (!serializable) {
```
patricia-causalens, line 61: Do we need any warnings if it is not serializable? Not sure when this would happen but wondering if it might make it easier to pinpoint issues

packages/dara-core/js/shared/interactivity/persistence.tsx
```diff
@@ -0,0 +1,307 @@
+import { mixed } from '@recoiljs/refine';
+import * as React from 'react';
```
patricia-causalens, line 2: In most places we have been importing what was needed instead of doing React.useSomething. Is this better?

patricia-causalens (approved):
LGTM, great work! :)

packages/dara-core/js/api/http.tsx
krzysztof-causalens: This just exists in case `toSerializable` returns null/undefined, not if the value is not serializable

packages/dara-core/js/shared/interactivity/persistence.tsx
krzysztof-causalens: Recently we've been doing both interchangeably, doesn't matter too much - personal preference. Both are fine and should be handled by bundlers to treeshake correctly.

tamara-causalens (approved):

PR #198, opened 2024-02-12
Fix: DOS-649 Catch errors thrown in the variable state provider

sam-causalens:
## Motivation and Context
Tracing a bug that caused all the pick variables to break I discovered this issue with the VariableStateProvider component.

## Implementation Description
* If errors are not caught here then a single crash takes down this stream handling completely and no more requests for variable values are processed.
* As these requests can sometimes involve reaching out to a backend then this code should be resilient to faults and not break the behaviour of the app for subsequent requests.

No

## How Has This Been Tested?
Tested Locally

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/shared/variable-state-provider/variable-state-provider.tsx
+++ b/packages/dara-core/js/shared/variable-state-provider/variable-state-provider.tsx
@@ -15,8 +15,15 @@ function VariableStateProvider(props: VariableStateProviderProps): JSX.Element {
 
     useEffect(() => {
         const sub = props.wsClient?.variableRequests$().subscribe((req) => {
-            const variableValue = getVariableState(req.message.variable);
-            props.wsClient.sendVariable(variableValue, req.message.__rchan);
+            // Catch any errors when fetching the variable value otherwise this takes down the stream and no further
+            // requests are processed.
+            try {
+                const variableValue = getVariableState(req.message.variable);
+                props.wsClient.sendVariable(variableValue, req.message.__rchan);
+            } catch (err) {
+                // eslint-disable-next-line no-console
+                console.warn(`Error when processing a getVariableValue request: ${err}`);
+            }
         });
 
         return () => {
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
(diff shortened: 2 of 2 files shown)


packages/dara-core/js/shared/variable-state-provider/variable-state-provider.tsx
```diff
@@ -15,8 +15,15 @@ function VariableStateProvider(props: VariableStateProviderProps): JSX.Element {
...
-            const variableValue = getVariableState(req.message.variable);
-            props.wsClient.sendVariable(variableValue, req.message.__rchan);
+            // Catch any errors when fetching the variable value otherwise this takes down the stream and no further
+            // requests are processed.
+            try {
+                const variableValue = getVariableState(req.message.variable);
+                props.wsClient.sendVariable(variableValue, req.message.__rchan);
+            } catch (err) {
+                // eslint-disable-next-line no-console
+                console.warn(`Error when processing a getVariableValue request: ${err}`);
```
krzysztof-causalens, line 25: In case of an error, should we send back an empty value or are we just relying on the backend timing out? What kind of error could happen here out of curiosity?

krzysztof-causalens (approved):

packages/dara-core/js/shared/variable-state-provider/variable-state-provider.tsx
sam-causalens: I was planning on letting it timeout given that we broadcast these request messages to all connected sessions for the user and didn't want this to quickly return a None value and not give another the chance to respond if it did actually have the value. 

The cause of the error here is actually the backend timing out (I think). Essentially we are ending up in a deadlock with the kernel as its calling get_current_value when executing a cell, which goes to the ui and ends up calling this logic which can trigger a call for a DerivedVariable. If that DerivedVariable is not cached then it ends up on the end of the kernel queue which is blocked by the first piece of code waiting for the current value and we deadlock until the get_current_value timeout breaks us out of it.  

Fixing the deadlock is more complicated than we really have time for now, so this patch at least lets the component recover and keep processing new requests going forwards even if this fails.

krzysztof-causalens: Right, I see. Fair enough, happy with the solution for the time being.

PR #204, opened 2024-02-15
Improvement: DO-2575: Disable UploadDropzone paste event by default

Roman-Kornev:
## Motivation and Context

Right now `UploadDropzone` includes an unconditional event handler for any `paste` event. This is fine for use in full-page apps, but presents a problem when it is used in larger pages with multiple components.

This PR adds an `enable_paste` prop that will control whether the `paste` event listener is enabled or not. By default, it is now disabled, requiring explicit activation.

## How Has This Been Tested?

Locally

## PR Checklist:


tamara-causalens (approved):

packages/dara-components/dara/components/common/dropzone.py
```diff
@@ -67,6 +69,7 @@ def __init__(
             - if a target is specified, can be used to customise how the `bytes` received are turned into a `DataFrame`
             - if a target is not specified, can be treated as a side effect function to run on the `bytes` received (to i.e. store on disk)
         :param on_drop: optional action triggered when a file is successfully uploaded
+        :param enable_paste: optional flag to enable pasting of text as files
```
patricia-causalens, line 72: Could we make this a bit clearer? What it does is that it stops us adding the paste event listener. Which means mostly that the user can't cmd/ctrl + V or paste via paste option on the mouse context menu. I think a bit more just saying what it enables and what it can be useful for/when to avoid briefly would be useful.

Roman-Kornev: I realize I cut a bit too much context when translating from JS side to Python. Expanded on the description a bit more. Would this be helpful?

patricia-causalens: Yes, looks good :)

krzysztof-causalens:
@Roman-Kornev could you fix CI and get this merged? Then we can cherry-pick to 1.6.x

tamara-causalens:
Hey @Roman-Kornev shall we tackle the tests to pass, is it good to merge then?

Roman-Kornev:
@krzysztof-causalens @tamara-causalens tests should pass automatically when dara-ui 1.6.1 is released

PR #215, opened 2024-02-19
Fix: DOS-727 address the disconnect of the websocket on subsequent disconnects

sam-causalens:
## Motivation and Context
People were still seeing their instances hang up and after some investigation I found that this was occuring on the second disconnect of the websocket whilst the first recovered fine. This was found to be a small mistake in the previous fix I implemented.

## Implementation Description
* The previous fix forgot to readd the close handler to the new socket, so another disconnect would leave the socket in the same state as before.

No

## How Has This Been Tested?
Tested in a local instance of studio where a button was added to simulate multiple disconnect/reconnects in a row.

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/api/websocket.tsx
+++ b/packages/dara-core/js/api/websocket.tsx
@@ -282,7 +282,7 @@ export class WebSocketClient implements WebSocketClientInterface {
                 }
             });
         });
-        // Rebind the close handler so this logic is re-applied on a subsequent disconnect
+        // Bind the close handler so the re-initialize logic is added every time
         this.socket.addEventListener('close', this.closeHandler);
     }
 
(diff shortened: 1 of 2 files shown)


packages/dara-core/js/api/websocket.tsx
```diff
@@ -302,6 +302,8 @@ export class WebSocketClient implements WebSocketClientInterface {
                 }
             });
         });
+        // Rebind the close handler so this logic is re-applied on a subsequent disconnect
+        this.socket.addEventListener('close', this.closeHandler);
```
krzysztof-causalens, line 306: Should we extract logic from the constructor and simply call it again here? Seems like the duplication is a bit brittle and can lead to more mistakes

krzysztof-causalens (approved):

PR #216, opened 2024-02-20
Feat: DO-2535 collaboration stage 2 - support storing unique value per user + FileBackend

krzysztof-causalens:
## Motivation and Context

This PR implements stage 2 of the collaboration feature.

This includes extending the `BackendStore` to support a `scope` parameter which can be set to `global` (default) or `user`.

In user mode, the read/write/delete methods pick up current user from the context variable and use it as the key to store data. This means current user can read/write/delete only their own data.

Added `get_all` method to list out all user data in that variable, as a dict of `{user: value}`. 

Added `FileBackend` as a persistent storage implementation for backend store.

## Implementation Description

Implemented `FileBackend` by pointing it at a location of a JSON file, which stores values keyed by user id (or just 'global'). The read/write operations are secured by an async RW lock to prevent data races.

Updated `BackendStore` to support `scope`. When scope=user, notify invokes a new method `send_message_to_user` in websocket manager instead of broadcast. This only sends notifications about the new value to channels belonging to the user who initiated the operation. 

Other changes:
- added `format-check` CI step and commands to enforce formatting
- added `tsc --noemit` to JS linting step to make spotting type errors easier

`aiorwlock` - simple read/write async lock implementation used for locking the file backend 

## How Has This Been Tested?

Unit tested and in a local app

## PR Checklist:

https://github.com/causalens/dara/assets/87647189/b57e1548-fad9-499a-bab2-86f0f08d418c


tamara-causalens (approved):
This looks great Krys!

sam-causalens (approved):
LGTM

packages/dara-core/dara/core/persistence.py
```diff
@@ -76,10 +142,47 @@ def dict(self, *args, **kwargs):
...
+
+    :param uid: unique identifier for this store; defaults to a random UUID
+    :param backend: the backend to use for storing data; defaults to an in-memory backend
+    :param scope: the scope for the store; if 'global' a single value is stored for all users,
+        if 'user' a value is stored per user
     """
 
     uid: str = Field(default_factory=lambda: str(uuid4()))
     backend: PersistenceBackend = Field(default_factory=InMemoryBackend, exclude=True)
+    scope: Literal['global', 'user'] = 'global'
```
patricia-causalens, line 154: For the type here, for example with cache we use CacheType being a enum but also accepting their string equivalent. Do we want to do the same here?

patricia-causalens (commented):
LGTM! Just a small question :)

packages/dara-core/dara/core/persistence.py
krzysztof-causalens: This is not equivalent to `CacheType` as we don't have a `session` option available. If you're asking whether we should introduce an enum for those, I think it's overkill for this simple of a case and strings should be enough

PR #224, opened 2024-02-28
Fix: DOS-764 Allow async startup functions to be registered

sam-causalens:
## Motivation and Context
* Needed an async startup function for an internal project

## Implementation Description
* Moves the execution of startup functions to be part of the app lifecycle so we can make allow async function. Detects coroutines being returned and then awaits them.

## How Has This Been Tested?
Unit tests added

## PR Checklist:


tamara-causalens (approved):
Great, thanks for the change! 

Note: This will need to be cherry picked to 1.6 as well.

krzysztof-causalens (approved):

patricia-causalens (approved):
LGTM

PR #228, opened 2024-03-04
Fix: DO-2651 settings .env file generation failure

krzysztof-causalens:
## Motivation and Context

Currently in Dara when an .env file is missing when attempting to read from environment variables, it attempts to generate one in the CWD. In some cases this directory is not writable. 

This PR changes the behaviour to instead fall back to using in-memory secrets rather than the `.env` file in addition to logging the error.

## Implementation Description

Try/except added on the env generation logic. In case of failure `_env_file=None` is passed to the `Settings` class as per [Pydantic docs](https://docs.pydantic.dev/1.10/usage/settings/#dotenv-env-support) to make it not attempt to load any files.

The Settings class already as defaults set up for crucial envs like jwt_secret.

## How Has This Been Tested?

Added a unit test to cover the scenario

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -5,6 +5,8 @@ title: Changelog
 ## NEXT
 
 -   Added the ability to pass an asynchronous function to `ConfigurationBuilder.on_startup(...)`.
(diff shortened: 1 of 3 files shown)


tamara-causalens (approved):

patricia-causalens (approved):
LGTM

PR #235, opened 2024-03-13
Feat: DO-2694 chat component dara

patricia-causalens:
## Motivation and Context

Adding a new component to aid collaboration within Dara apps.

## Implementation Description

Connected the dara-ui Chat component so that  it can be called on Dara side. 
Added in the docstrings how this could be used in collaboration with the new BackendStore to promote collaboration.

Added some logic to switch between the chat button and chat interface.

Will update dara-ui once that PR is approved and update it here

## How Has This Been Tested?

Locally on a demo app

## PR Checklist:

multi user chat


packages/dara-components/dara/components/smart/chat/chat.py
```diff
@@ -0,0 +1,63 @@
...
+
+class Chat(StyledComponentInstance):
+    """
+    A Chat component which can be added anywhere in your page. When added a chat button will appear in the bottom right
+    of your page and when clicked a chat sidebar will appear. This can be added on a page by page basis, with the chat
+    state being store in a Variable.
+
+    If you would like for the Variable's state to be persistent between restarts, and shared across multiple app users,
+    you can use a BackendStore with your prefered backend. For example, to store the chat messages in a json file, you
+    can use the FileBackend, as showcased below:
```
krzysztof-causalens, line 30: I think this could be made clearer:
1. You *have* to use a BackendStore for this to behave correctly, otherwise each user has their own ephemeral chat state which is quite pointless. Side note: perhaps we should validate the value variable has a BackendStore attached?
2. You can choose to use a persistent backend rather than an ephemeral one (i.e. currently File rather than InMemory) for it to persist across restarts.

Your current wording suggests you have to use a backend store for persistence between restarts but that's not quite correct,s ee above.

packages/dara-components/js/smart/chat/chat.tsx
```diff
@@ -0,0 +1,89 @@
...
+ */
+function Chat(props: ChatProps): JSX.Element {
+    const [style, css] = useComponentStyles(props);
+    const [value, setValue] = useVariable(props.value);
+    const [showChat, setShowChat] = React.useState(false);
+
+    return (
+        <ChatWrapper>
+            {showChat && (
+                <StyledChat
```
krzysztof-causalens, line 59: A minor UI improvement would be to have a brief transition for when this appears on the page but not necessary especially for v1

packages/dara-components/js/smart/chat/chat.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    className={props.className}
+                    onClose={() => setShowChat(false)}
+                    onUpdate={setValue}
+                    style={style}
+                    value={value}
+                />
+            )}
+            {!showChat && (
+                <ChatButton onClick={() => setShowChat(true)}>
+                    <svg fill="none" height="32" viewBox="0 0 32 32" width="32" xmlns="http://www.w3.org/2000/svg">
```
krzysztof-causalens, line 70: This should probably use our theme colors so it works with dark mode / other themes

packages/dara-components/js/smart/chat/chat.tsx
patricia-causalens: Good point, have added a reminder to the spec so we can do this :)

packages/dara-components/dara/components/smart/chat/chat.py
patricia-causalens: Updated, let me know if you think this is better, have also added a validation to see if a store is attached

packages/dara-components/dara/components/smart/chat/chat.py
```diff
@@ -61,3 +63,14 @@ class Chat(StyledComponentInstance):
...
         extra = 'forbid'
+
+    @validator('value')
+    @classmethod
+    def validate_vaeiable(cls, value):
+        """
+        Validate that the Variable has a store attached to it.
+        """
+        if value.store is None:
+            raise ValueError('A Chat value variable must have a store attached to it')
```
krzysztof-causalens, line 74: In the future we might add stores which are not `BackendStore`, e.g. for local storage or URL storage. This should check that the store is not None and an instance of BackendStore

patricia-causalens: Added

krzysztof-causalens (approved):

PR #242, opened 2024-03-25
Fix: DO-2781: Dara serialization: MultiIndex on a pandas.DataFrame

AidaZheng:
## Motivation and Context
This bug happens when a `Variable` have a DataFrame and the DataFrame is multi-indexed.

The root reason is that when a df have multiple index, `to_dict()` will use tuple type as key.
However, tuple is not jsonable, hence, jsonable_enocder will turn it into a list. List can't not be used as a key of dict.

## Implementation Description
To avoid this, when serializing a df, will iterate the dict. If it has a tuple type key, will turn it into a string with a __tuple__ prefix. So the whole dict is jsonable.

However, this require user to  deserialise it, by doing
```
def _tuple_key_deserialize(d):
    encoded_dict = {}
    for key,value in d.items():
        encoded_key = ast.literal_eval(key[9:])if isinstance(key,str)and key.startswith('__tuple__') else key
        encoded_value = _tuple_key_deserialize(value) if isinstance(value,dict) else value
        encoded_dict[encoded_key] = encoded_value
    return encoded_dict 
preds_df = pandas.DataFrame(_tuple_key_deserialize(d))
```
Happy to discuss if there's better solution.

No

## How Has This Been Tested?
Locally tested

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -15,7 +15,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 # pylint: disable=unnecessary-lambda
-import ast
 from inspect import Parameter, isclass
 from typing import (
     Any,
@@ -96,7 +95,7 @@ def _tuple_key_deserialize(d):
     """
     encoded_dict = {}
     for key, value in d.items():
-        encoded_key = ast.literal_eval(key[9:]) if isinstance(key, str) and key.startswith('__tuple__') else key
+        encoded_key = tuple(key[10:-1].split(',')) if isinstance(key, str) and key.startswith('__tuple__') else key
         encoded_value = _tuple_key_deserialize(value) if isinstance(value, dict) else value
         encoded_dict[encoded_key] = encoded_value
     return encoded_dict
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -15,6 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 # pylint: disable=unnecessary-lambda
+import ast
 from inspect import Parameter, isclass
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -134,7 +143,7 @@ def _get_pandas_array_encoder(array_type: Type[Any], dtype: Any, raise_: bool =
     pandas.Index: Encoder(serialize=lambda x: x.to_list(), deserialize=lambda x: pandas.Index(x)),
     pandas.Timestamp: Encoder(serialize=lambda x: x.isoformat(), deserialize=lambda x: pandas.Timestamp(x)),
     pandas.DataFrame: Encoder(
-        serialize=lambda x: jsonable_encoder(x.to_dict(orient='dict')),
+        serialize=lambda x: jsonable_encoder(_tuple_key_serialize(x.to_dict(orient='dict'))),
         deserialize=lambda x: x if isinstance(x, pandas.DataFrame) else _not_implemented(x, pandas.DataFrame),
```
krzysztof-causalens, line 147: Can we modify the default deserializer to detect the `__tuple__` prefixed keys and automatically turn them into a multi index or is that not possible without knowing the types? We could encode more information in the serialized tuple if need be

packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -71,6 +71,15 @@ def _get_pandas_array_encoder(array_type: Type[Any], dtype: Any, raise_: bool =
     )
 
 
+def _tuple_key_serialize(d):
```
krzysztof-causalens, line 74: Can we add a docstring explaining what and why we're doing here for future reference

krzysztof-causalens (commented):
Looks good, couple questions

packages/dara-core/dara/core/internal/encoder_registry.py
AidaZheng: I guess I can add a deserializer for df. If a user add an annotation type DataFrame, I can check if the value is a dict type, if it is, turn the dict into a multi indexed df.

However, if a user doesn't add an annotation type, or the type is List[DataFrame]/Dict[str,DataFrame], the deserializer won't work.

krzysztof-causalens: Yeah the deserializer *only* runs when the user annotates it as a DF, which I think is fine.

packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -80,6 +88,34 @@ def _tuple_key_serialize(d):
...
 
+def _tuple_key_deserialize(d):
+    """
+    A helper function to recursively check if a dict have a key with "__tuple__" prefix and turn it back to tuple
+
+    :param d: the dict where tuple key to be deserialized
+    """
+    encoded_dict = {}
+    for key, value in d.items():
+        encoded_key = ast.literal_eval(key[9:]) if isinstance(key, str) and key.startswith('__tuple__') else key
```
krzysztof-causalens, line 99: This is fairly risky as we're evaling data sent from the browser directly. 
According to [the docs](https://docs.python.org/3/library/ast.html#ast.literal_eval):
> This function had been documented as “safe” in the past without defining what that meant. That was misleading. This is specifically designed not to execute Python code, unlike the more general [eval()](https://docs.python.org/3/library/functions.html#eval). There is no namespace, no name lookups, or ability to call out. But it is not free from attack: A relatively small input can lead to memory exhaustion or to C stack exhaustion, crashing the process. There is also the possibility for excessive CPU consumption denial of service on some inputs. Calling it on untrusted data is thus not recommended.

Can we think of a different solution without using `eval`?

krzysztof-causalens (changes requested):

krzysztof-causalens (approved):

PR #248, opened 2024-04-04
Fix: DOS-887 Add chunking of websocket messages

sam-causalens:
## Motivation and Context
* An internal project needs to send back a bunch of data in reponse to a websocket message and it doesn't all fit within a single response, so this change adds the ability to chunk responses.

## Implementation Description
* Adds the ability to pass a chunk_count parameter along with DaraClientMessages to denote that there are multiple parts to this message that will be sent separately.
* On the backend side we then wait for the specified number of messages to be received before setting the event to unblock the sendAndWait handler. We make no attempt to recombine the messages on the backend at this time as we don't know what format the response will be in on the framework side. It's left up to the calling code to know when chunked responses might be returned and handle the merging appropriately.
* A new api `sendMessage` has been added to the frontend's websocket client to allow non variable messages to be sent back with a chunkCount set.
* Lastly, there was an odd bug in the WebsocketHandler in that all handlers ended up using the same object for pending_responses as we were relying on the default value which is defined only once rather than on each instantiation of the class.

No

## How Has This Been Tested?
Unit tests to be added (already written on the internal project, just need transferring here)

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/internal/websocket.py
+++ b/packages/dara-core/dara/core/internal/websocket.py
@@ -40,10 +40,13 @@ from dara.core.logging import dev_logger, eng_logger
 class DaraClientMessage(BaseModel):
     """
     Represents a message sent by Dara internals from the frontend to the backend.
+
+    An optional chunk_count field can be used to indicate that a message is chunked and what number of messages to expect
     """
 
     type: Literal['message'] = Field(default='message', const=True)
     channel: str
+    chunk_count: Optional[int] = None
     message: Any
 
 
@@ -132,7 +135,7 @@ class WebSocketHandler:
     Stream containing messages to send to the client.
     """
 
-    pending_responses: Dict[str, Tuple[Event, Optional[Any]]] = {}
+    pending_responses: Dict[str, Tuple[Event, Optional[Any]]]
     """
     A map of pending responses from the client. The key is the message ID and the value is a tuple of the event to
     notify when the response is received and the response data.
@@ -146,6 +149,7 @@ class WebSocketHandler:
         self.channel_id = channel_id
         self.send_stream = send_stream
         self.receive_stream = receive_stream
+        self.pending_responses = {}
 
     async def send_message(self, message: ServerMessage):
         """
@@ -168,11 +172,23 @@ class WebSocketHandler:
             # If the message has a channel ID, it's a response to a previous message
             if message_id:
                 if message_id in self.pending_responses:
-                    event, _ = self.pending_responses[message_id]
+                    event, existing_messages = self.pending_responses[message_id]
 
-                    # Store the response and set the event to notify the waiting coroutine
-                    self.pending_responses[message_id] = (event, message.message)
-                    event.set()
+                    # If the response is chunked then collect the messages in pending responses
+                    if message.chunk_count is not None:
+                        if existing_messages is not None and isinstance(existing_messages, list):
+                            existing_messages.append(message.message)
+                        else:
+                            existing_messages = [message.message]
+                            self.pending_responses[message_id] = (event, existing_messages)
+
+                        # If all chunks have been received, set the event to notify the waiting coroutine
+                        if len(existing_messages) == message.chunk_count:
+                            event.set()
+                    else:
+                        # Store the response and set the event to notify the waiting coroutine
+                        self.pending_responses[message_id] = (event, message.message)
+                        event.set()
 
         if message.type == 'custom':
             # import required internals
@@ -195,7 +211,7 @@ class WebSocketHandler:
                 eng_logger.error(f'No handler found for custom message kind {kind}', e)
                 return
 
-    async def send_and_wait(self, message: ServerMessage) -> Optional[ClientMessage]:
+    async def send_and_wait(self, message: ServerMessage) -> Optional[Any]:
         """
         Send a message to the client and return the client's response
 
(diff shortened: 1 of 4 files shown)


packages/dara-core/dara/core/internal/websocket.py
```diff
@@ -132,7 +135,7 @@ class WebSocketHandler:
     Stream containing messages to send to the client.
     """
 
-    pending_responses: Dict[str, Tuple[Event, Optional[Any]]] = {}
```
krzysztof-causalens, line 135: Oops, nice catch!

packages/dara-core/dara/core/internal/websocket.py
```diff
@@ -168,11 +172,23 @@ async def process_client_message(self, message: ClientMessage):
...
-                    event.set()
+                        # If all chunks have been received, set the event to notify the waiting coroutine
+                        if len(existing_messages) == message.chunk_count:
+                            event.set()
+                    else:
+                        # Store the response and set the event to notify the waiting coroutine
+                        self.pending_responses[message_id] = (event, message.message)
+                        event.set()
 
         if message.type == 'custom':
```
krzysztof-causalens, line 193: Do we want the same chunking logic for custom messages, i.e. wait for all chunks and combine them before passing into user handler? Otherwise it will only work for internal messages

packages/dara-core/dara/core/internal/websocket.py
```diff
@@ -195,7 +211,7 @@ async def process_client_message(self, message: ClientMessage):
                 eng_logger.error(f'No handler found for custom message kind {kind}', e)
                 return
 
-    async def send_and_wait(self, message: ServerMessage) -> Optional[ClientMessage]:
+    async def send_and_wait(self, message: ServerMessage) -> Optional[Any]:
```
krzysztof-causalens, line 214: What's the reason to use Any here? I thought ClientMessage was pretty loosely-defined anyway

packages/dara-core/js/api/websocket.tsx
```diff
@@ -433,10 +433,22 @@ export class WebSocketClient implements WebSocketClientInterface {
...
+     * @param value variable value to send
+     * @param channel return channel to identify the message
+     * @param chunkCount total number of chunks this message has been split into
+     */
+    sendMessage(value: any, channel: string, chunkCount?: number): void {
         if (this.socket.readyState === WebSocket.OPEN) {
             this.socket.send(
                 JSON.stringify({
                     channel,
+                    chunk_count: chunkCount ?? null,
```
krzysztof-causalens, line 451: I think we also want this added to `sendCustomMessage` below?

packages/dara-core/dara/core/internal/websocket.py
```diff
@@ -40,10 +40,13 @@
 class DaraClientMessage(BaseModel):
     """
     Represents a message sent by Dara internals from the frontend to the backend.
+
+    An optional chunk_count field can be used to indicate that a message is chunked and what number of messages to expect
     """
 
     type: Literal['message'] = Field(default='message', const=True)
     channel: str
+    chunk_count: Optional[int] = None
```
krzysztof-causalens, line 49: Would also be needed in CustomClientMessage

krzysztof-causalens (commented):
Looks good in general, I think what you've implemented is enough - AFAIK message order is guaranteed in websockets since it's just TCP under the hood. 

My main point below is that I think we need the same logic for custom ws messages, not just internal ones

packages/dara-core/dara/core/internal/websocket.py
sam-causalens: This is because we don't actually return the ClientMessage object, we return ClientMessage.message which has an Any type. This is just matching the type to what we actually return. Is there a way to do `ClientMessage.message` as the type? I tried it that way and as `ClientMessage['message']` but both broke in different ways...

packages/dara-core/dara/core/internal/websocket.py
sam-causalens: Happy to add for customer messages as well, just wanted to verify the approach before making too many changes.

packages/dara-core/dara/core/internal/websocket.py
krzysztof-causalens: Ah I see. Yeah there isn't a way to do that in Python, I don't think 😅 Happy to keep as is

patricia-causalens (approved):
Looks good! Could we just add an entry to the changelog?

packages/dara-core/tests/python/test_websocket.py
```diff
@@ -36,6 +47,76 @@ class LocalJsComponent(ComponentInstance):
 os.environ['DARA_DOCKER_MODE'] = 'TRUE'
 
 
+async def _await_for_condition(fn: Callable[..., Any]):
```
krzysztof-causalens, line 50: We have a few similar utils in `tests/python/utils.py`, like `wait_for`

packages/dara-core/tests/python/test_websocket.py
```diff
@@ -36,6 +47,76 @@ class LocalJsComponent(ComponentInstance):
...
+        nonlocal result
+        result = await ws_handler.send_and_wait(DaraServerMessage(message=ServerMessagePayload()))
+
+    # Run the two tasks concurrently
+    async with anyio.create_task_group() as tg:
+        tg.start_soon(_send_task, ws_handler, ['hello', 'world', '!!!'])
+        tg.start_soon(get_task)
+
+    # Verify that a single doc was returned
+    assert result == ['hello', 'world', '!!!']
```
krzysztof-causalens, line 117: I think the comment was meant to say multiple rather than single?

krzysztof-causalens (approved):
LGTM. 
For extra context for others, we've discussed the issue offline and decided to only implement chunking for internal messages - the current intended use case does not use custom msgs and it would take some work to figure out a better approach to support chunking for custom messages.

PR #256, opened 2024-04-05
Fix: DO-2781 Fix multiple index df deserialization

AidaZheng:
## Motivation and Context
The previous deserialise method will turn tuple to ["'mean'", "'median'"] instead of  ["mean", "median"]

## Implementation Description
For each string, will split by `,` first, then check it the result start and end with `'` or `"`, if so, remove it. 

No 

## How Has This Been Tested?
Integration tested

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -106,7 +106,6 @@ def _tuple_key_deserialize(d):
         else:
             encoded_key = key
 
-        # encoded_key = tuple(key[10:-1].split(',')) if isinstance(key, str) and key.startswith('__tuple__') else key
         encoded_value = _tuple_key_deserialize(value) if isinstance(value, dict) else value
         encoded_dict[encoded_key] = encoded_value
     return encoded_dict
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -95,7 +95,18 @@ def _tuple_key_deserialize(d):
     """
     encoded_dict = {}
     for key, value in d.items():
-        encoded_key = tuple(key[10:-1].split(',')) if isinstance(key, str) and key.startswith('__tuple__') else key
+        if isinstance(key, str) and key.startswith('__tuple__'):
+            key_list = []
+            for each in key[10:-1].split(', '):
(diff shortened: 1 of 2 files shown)


packages/dara-core/dara/core/internal/encoder_registry.py
```diff
@@ -95,7 +95,18 @@ def _tuple_key_deserialize(d):
...
+            for each in key[10:-1].split(', '):
+                if (each.startswith("'") and each.endswith(("'"))) or (each.startswith('"') and each.endswith(('"'))):
+                    key_list.append(each[1:-1])
+                else:
+                    key_list.append(each)
+            encoded_key = encoded_key = tuple(key_list)
+        else:
+            encoded_key = key
+
+        # encoded_key = tuple(key[10:-1].split(',')) if isinstance(key, str) and key.startswith('__tuple__') else key
```
krzysztof-causalens, line 109: Leftover comment?

krzysztof-causalens (commented):
Looks good, can we add a test case to ensure the tuple serialization->deserialization is correct?

AidaZheng:
> Looks good, can we add a test case to ensure the tuple serialization->deserialization is correct?

I have a test to cover serialization->deserialization already, but it didn't have a `assert` before, hence, the error is not been find out. But the assert is added and it pass the test :)

packages/dara-core/tests/python/test_encoders.py
```diff
@@ -93,10 +93,11 @@ def test_serialization_deserialization(type_, value, raise_):
     deserialized = encoder['deserialize'](json.loads(serialized))
 
     # If it's a pandas type use built-in equality
-    if isinstance(value, pandas.Series) or isinstance(value, pandas.Index) :
+    if value is df_with_datetime_index:
+        # DF with datetime index like 2023-10-04 will be encoded and decoded to 2023-10-04T00:00:00
+        pass
```
krzysztof-causalens, line 98: Why is there no assertion in this case?

krzysztof-causalens (approved):
Makes sense, one more question

claudiawestby (approved):
this works for my test case - thank you!

packages/dara-core/tests/python/test_encoders.py
AidaZheng: If it has a timestamp like `2023-10-04` in df, our encoder will turn it into `2023-10-04T00:00:00`. And the decoder can't turn it back to `2023-10-04`, as it doesn't know the orginal timestamp format. The decoder only knows it as a df.

PR #261, opened 2024-04-16
Feat: DO-2798: chat component send notifications

patricia-causalens:
## Motivation and Context

Update chat so that it can send notifications to defined endpoint when a message is sent by a user.

## Implementation Description

When a meesage is sent we send a message to chat/messages endpoint on dara. Have added so an user can set a ChatConfig and in it have a callback for what happens when this endpoint is hit. 
If the config is not set then the component works as normal.

None

## How Has This Been Tested?

Locally by seeing that the request was sent with the expected payload.

## PR Checklist:

notification


packages/dara-components/js/smart/chat/chat.tsx
```diff
@@ -80,14 +124,29 @@ function Chat(props: ChatProps): JSX.Element {
     const [showChat, setShowChat] = React.useState(false);
     const theme = useTheme();
 
+    const onUpdate = (newValue: Message[]): void => {
+        // If the new value is longer than the old value, we can assume that a new message was added
+        if (newValue.length > value.length) {
+            const newMessage = newValue[newValue.length - 1];
+            const users = getAllUsersInChat(newValue);
```
patricia-causalens, line 131: Is it better to have a copy of all the users in a state in the component or calculate it whenever a message is sent?

krzysztof-causalens (changes requested):
Looks good, but needs a little bit more work to be able to wire up the notifications correctly

packages/dara-components/js/smart/chat/chat.tsx
krzysztof-causalens: I think that's premature optimization, realistically how often will this be invoked? You're not really gaining enough to warrant the increased complexity in code. 
Let's start with how it is right now, if we notice it causing issues and measure that this is a bottleneck then we can revisit.

packages/dara-components/dara/components/smart/chat/config.py
```diff
@@ -0,0 +1,23 @@
...
+    Chat configuration object.
+
+    :param on_new_message: Callback function to be called when a new message is sent in the chat
+    """
+
+    on_new_message: Callable[[NewMessageBody], None]
+
+    @classmethod
+    def default(cls):
+        dev_logger.warning('Using default configuration for ChatConfig. Please provide a valid configuration.')
```
krzysztof-causalens, line 22: Do we want to log a warning here? It's completely fine to not use notifications for the Chat I think?

packages/dara-components/dara/components/smart/chat/endpoints.py
```diff
@@ -0,0 +1,16 @@
...
+Copyright (c) 2023 by Impulse Innovations Ltd. Private and confidential. Part of the causaLens product.
+"""
+
+import mimetypes
+
+from dara.core.http import post
+from dara.components.smart.chat.config import ChatConfig
+from dara.components.smart.chat.types import NewMessageBody
+
+mimetypes.init()
```
krzysztof-causalens, line 11: What is this needed for here?

packages/dara-components/js/smart/chat/chat.tsx
```diff
@@ -96,14 +99,19 @@ function getAllUsersInChat(chat: Message[]): UiUserData[] {
...
-        });
+        const res = await request(
+            '/api/chat/messages',
+            {
+                body: JSON.stringify(payload),
+                method: HTTP_METHOD.POST,
+            },
+            extras
+        );
+        await handleAuthErrors(res);
```
krzysztof-causalens, line 114: Should provide `true` as the second argument to redirect to login on an error rather than to the error page

packages/dara-components/dara/components/smart/chat/endpoints.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@post(f'/chat/messages')
+def on_new_message(chat_config: ChatConfig, body: NewMessageBody):
+    return chat_config.on_new_message(body)
```
krzysztof-causalens, line 16: Should probably wrap it in `run_user_handler` to make sure it can be async or sync.

packages/dara-components/changelog.md
```diff
@@ -6,6 +6,7 @@ title: Changelog
 
 -   `Chat` component now shows the user who has written a message
 -   `Chat` component now displays its messages content as `Markdown`.
+-   `Chat` sends a notification when a new message is addded to it.
```
krzysztof-causalens, line 9: That's not entirely true since it requires configuring, should we expand about it here a bit? Maybe a small snippet on how a user can configure what happens to new messages and that it can be used to e.g. send notifications

patricia-causalens: Forgot to update changelog after changes, good catch!

packages/dara-components/dara/components/smart/chat/endpoints.py
patricia-causalens: It is not, leftover from a copy paste, removed

packages/dara-components/dara/components/smart/chat/endpoints.py
patricia-causalens: Good point, added

krzysztof-causalens (approved):

PR #262, opened 2024-04-16
Fix: DO-2868 redirect to login on auth errors in internal requests

krzysztof-causalens:
## Motivation and Context

It's been reported that there are issues in an internal product where features (using e.g. Dara actions) stop working due to login expiring but the user is not kicked out to the login page immediately.

After investigating, I found that the issue was due to the action request not handling auth errors properly. With the fix upon an auth error the user will be immediately redirected to the login page. 

Looking through internal requests made, that piece of logic was missing for few other requests so I've also fixed those for consistency.

## Implementation Description

Added `handleAuthErrors` call to each usage of `request` within Dara internals (derived variables, py_components, actions etc). This catches auth errors thrown by Dara and redirects to login page if one is found.

## How Has This Been Tested?

Verified actions were the issue in the internal product by modifying a local JWT to be expired
Modified local Dara code in the product to include the fix and verified that causes an immediate redirection to login page.

## PR Checklist:

### DIFF
--- a/packages/dara-components/js/common/dropzone/dropzone.tsx
+++ b/packages/dara-components/js/common/dropzone/dropzone.tsx
@@ -7,6 +7,7 @@ import {
     DefaultFallback,
     RequestExtras,
     StyledComponentProps,
+    handleAuthErrors,
     injectCss,
     request,
     useAction,
@@ -61,6 +62,7 @@ async function uploadFileToExtension(
         },
         extras
     );
+    await handleAuthErrors(res, true);
     await validateResponse(res, `Failed to upload file: ${file.name}`);
     const result: { [k: string]: any } = await res.json();
     return { newStatus: result.status };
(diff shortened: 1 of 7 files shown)


krzysztof-causalens:
> We do have validateResponse without the auth for example in tasks and for getting the user, just so I understand better, what defined when we should check the auth as well?

I think we should check for auth errors in all requests to authenticated endpoints, which are most of them except for a few. Let me know if you see ones I've missed

patricia-causalens (approved):
LGTM

patricia-causalens:
> > We do have validateResponse without the auth for example in tasks and for getting the user, just so I understand better, what defined when we should check the auth as well?
> 
> I think we should check for auth errors in all requests to authenticated endpoints, which are most of them except for a few. Let me know if you see ones I've missed

Makes sense, had a look and couldn't find any others

tamara-causalens (approved):

PR #266, opened 2024-04-18
Fix: DOS-913 Websocket Reconnection Refactor

sam-causalens:
## Motivation and Context
* The existing logic for retrying when the websocket connection fails lead to it looping forever when trying to reconnect due to faulty logic that lead to multiple sockets being created and the number of attempts always being 0 so never hitting the max count.
* My previous changes had also not gone far enough in terms of updating the global socket variable that left this code in an odd state after a reconnection. For example, new connections did not get a proper heartbeat setup as this was done on the global level.

## Implementation Description
* To address this, I've moved all the socket connection logic to the initialize function, this is now where the socket is created and has all it's listeners attached. Calling this function will create a brand new socket everytime with all the correct handlers rather than the subset that was left after my previous changes.
* The logic around the onClose event of the websocket has been moved into the class itself and is no longer recursive. The reconnectCount is tracked on the class itself which makes it consistent across retries so the attempts correctly stop after 10seconds.
* The onClose logic now uses the revamped initialize function to ensure that the socket is setup in the exact same way on reconnect as it was before.
* The timeout is now properly respected between attempts and does not lead to an effective infinite loop of retry requests

jest-websocket-mock

## How Has This Been Tested?
* A suite of unit tests around the client have been added using `jest-websocket-mock` that test all the reconnection and initialization logic work correctly.

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/api/websocket.tsx
+++ b/packages/dara-core/js/api/websocket.tsx
@@ -182,43 +182,43 @@ export class WebSocketClient implements WebSocketClientInterface {
 
     closeHandler: () => void;
 
-    _maxAttempts: number;
+    maxAttempts: number;
 
-    _pingInterval: NodeJS.Timeout;
+    #pingInterval: NodeJS.Timeout;
 
-    _socketUrl: string;
+    #socketUrl: string;
 
-    _reconnectCount: number;
+    #reconnectCount: number;
 
     constructor(_socketUrl: string, _token: string, _liveReload = false) {
         this.token = _token;
         this.liveReload = _liveReload;
         this.messages$ = new Subject();
         this.closeHandler = this.onClose.bind(this);
-        this._socketUrl = _socketUrl;
-        this._reconnectCount = 0;
-        this._pingInterval = null;
-        this._maxAttempts = maxAttempts;
+        this.maxAttempts = maxAttempts;
+        this.#socketUrl = _socketUrl;
+        this.#reconnectCount = 0;
+        this.#pingInterval = null;
 
         // Lastly call initialize to setup the socket properly
         this.socket = this.initialize();
     }
 
-    initialize(): WebSocket {
+    initialize(isReconnect = false): WebSocket {
         // Create the underlying socket instance from the url and token
-        const url = new URL(this._socketUrl);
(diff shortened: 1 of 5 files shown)


packages/dara-core/js/api/websocket.tsx
```diff
@@ -258,48 +182,106 @@ export class WebSocketClient implements WebSocketClientInterface {
 
     closeHandler: () => void;
 
-    constructor(_socket: WebSocket, _token: string, _liveReload = false) {
-        this.socket = _socket;
+    _maxAttempts: number;
```
krzysztof-causalens, line 185: For private variables in JS you can use the [# prefix](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties)

packages/dara-core/js/api/websocket.tsx
```diff
@@ -258,48 +182,106 @@ export class WebSocketClient implements WebSocketClientInterface {
...
+        this._reconnectCount = 0;
+        this._pingInterval = null;
+        this._maxAttempts = maxAttempts;
+
+        // Lastly call initialize to setup the socket properly
+        this.socket = this.initialize();
     }
 
-    initialize(): void {
+    initialize(): WebSocket {
```
krzysztof-causalens, line 207: I might be missing something but I think this implementation doesn't implement liveReload at all?
Could we add a test case to ensure `location.reload` is invoked on reconnect if `liveReload` is true?

krzysztof-causalens (changes requested):
Looks good, nice find. My only issue is that the new implementation is missing the live reload feature as far as I can tell

packages/dara-core/js/api/websocket.tsx
sam-causalens: Ah, good catch. Sorry I took that out for simplicity when testing everything but forgot to put it back. I'll readd and add a test.

sam-causalens: Should be sorted now, let me know if you can think of any issues.

krzysztof-causalens (approved):

patricia-causalens (approved):
LGTM

PR #272, opened 2024-04-22
Improvement: DO-2850 attach chat button as overlay within the component instance

patricia-causalens:
## Motivation and Context

In internal product user wanted to add multiple chats to the same page

## Implementation Description

Changed so that the chat button that opens the thread now is set with relative to the parent. This allows the user to see multiple chats on the same page.

Since the threads are slightly transparent as per design, this meant that if multiple were open they would overlap, so one addition was for based on the number of threads on the page to set the latest opened to have a z-index higher than the others present.

If there were multiple threads we also add a background color so that chats below current are not visible.

## PR Checklist:

multi chat

### DIFF
--- a/packages/dara-components/js/smart/chat/chat.tsx
+++ b/packages/dara-components/js/smart/chat/chat.tsx
@@ -163,6 +163,7 @@ function Chat(props: ChatProps): JSX.Element {
 
     const [showChat, setShowChat] = React.useState(false);
     const [areThereOtherChats, setAreThereOtherChats] = React.useState(false);
+    const [zIndex, setZIndex] = React.useState(998);
 
     const user = useUser();
     const theme = useTheme();
@@ -186,6 +187,13 @@ function Chat(props: ChatProps): JSX.Element {
         setValue(newValue);
     };
 
+    const onClickChatButton = (): void => {
+        setShowChat(true);
+        setZIndex(getHighestZIndex('.chatThread') + 1);
+        // we need to update here so that if the chat button is clicked again when it is already open, the background color is updated
+        setAreThereOtherChats(checkMoreThanOneRenderedElement('.chatThread'));
+    };
+
     React.useLayoutEffect(() => {
         setAreThereOtherChats(checkMoreThanOneRenderedElement('.chatThread'));
     }, [showChat]);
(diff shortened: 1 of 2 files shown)


packages/dara-components/js/smart/chat/chat.tsx
```diff
@@ -149,60 +187,64 @@ function Chat(props: ChatProps): JSX.Element {
...
-                            d="M18.5981 26.1641L16 30.6641L13.4019 26.1641L18.5981 26.1641Z"
-                            fill={theme.colors.background}
-                            stroke={theme.colors.background}
-                        />
-                        <path d="M16 28.3359L13.4019 23.8359L18.5981 23.8359L16 28.3359Z" fill="none" />
-                    </svg>
-                </ChatButton>
+                // we set the z-index so that the latest chat thread opened is always on top, and if there is a chat thread open, we set the background color so that the transparency does not show the thread behind
+                <ThreadWrapper
+                    className="chatThread"
```
krzysztof-causalens, line 206: class names should be kebab-case

tamara-causalens (approved):
Looks great @patricia-causalens!

packages/dara-components/js/smart/chat/chat.tsx
```diff
@@ -30,26 +30,32 @@ interface MessageNotificationPayload {
...
+
+    border-radius: 0.4rem;
 `;
 
 const ChatButton = styled.button`
-    width: 3.5rem;
-    height: 3.5rem;
-    margin: 0 2rem 2rem 0;
-    padding-top: 0.5rem;
+    position: absolute;
```
krzysztof-causalens, line 46: Do all our layout components have `position: relative` for this to attach to?

krzysztof-causalens (approved):

packages/dara-components/js/smart/chat/chat.tsx
patricia-causalens: Yes, in the definition of the class for LayoutComponent we set the position to relative for all of them

packages/dara-components/js/smart/chat/chat.tsx
patricia-causalens: Oops had it as an id before and forgot to change conventions, fixed :)

PR #273, opened 2024-04-22
Fix: DO-2862 bokeh plotly change background to be transparent by default

patricia-causalens:
## Motivation and Context

When using Plotly with different components behind the background color of the plot didn't always match that of the parent

## Implementation Description

Updated to set the background color to `rgba(255,255,255,0)`, this is because `transparent` keyword is not considered a valid value for either theme.

None

## How Has This Been Tested?

Locally in light and dark theme and updating the parent background color

## PR Checklist:


krzysztof-causalens (approved):

tamara-causalens (approved):
Nice!

PR #274, opened 2024-04-22
Improvement: DO-2837 dara to use dara UI markdown component

patricia-causalens:
## Motivation and Context

To use dara-ui Markdown component, as to avoid duplicates across the product

## Implementation Description

Updated to use dara-ui Markdown component

None

## How Has This Been Tested?

Locally

## PR Checklist:

### DIFF
--- a/packages/dara-components/package.json
+++ b/packages/dara-components/package.json
@@ -72,12 +72,10 @@
     "prism-react-renderer": "^1.2.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-markdown": "^9.0.1",
     "react-plotly.js": "~2.6.0",
     "react-virtualized-auto-sizer": "1.0.6",
     "recoil": "^0.7.7",
-    "rehype-raw": "^6.1.1",
-    "remark-gfm": "^4.0.0"
+    "rehype-raw": "^6.1.1"
   },
   "files": [
     "dist"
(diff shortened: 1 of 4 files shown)


packages/dara-components/package.json
```diff
@@ -72,17 +72,15 @@
     "prism-react-renderer": "^1.2.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "react-markdown": "^9.0.1",
     "react-plotly.js": "~2.6.0",
     "react-virtualized-auto-sizer": "1.0.6",
     "recoil": "^0.7.7",
-    "rehype-raw": "^6.1.1",
-    "remark-gfm": "^4.0.0"
+    "rehype-raw": "^6.1.1"
```
krzysztof-causalens, line 78: I don't think we need the plugin as a direct dependency anymore?

krzysztof-causalens (approved):

packages/dara-components/package.json
patricia-causalens: We still need rehypeRaw plugin as that is not added on dara-ui side, but the gfm plugin was removed

tamara-causalens (approved):
lovely!

packages/dara-components/package.json
krzysztof-causalens: Right, makes sense

PR #277, opened 2024-04-26
Fix: DO-2938 DF deserialise with default index

AidaZheng:
## Motivation and Context
deserializing a df from a dict will create a array like index which is different from the original range type

## Implementation Description
Add a index type info when serialising. 

No

## How Has This Been Tested?
Unit test

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/internal/encoder_registry.py
+++ b/packages/dara-core/dara/core/internal/encoder_registry.py
@@ -111,6 +111,15 @@ def _tuple_key_deserialize(d):
     return encoded_dict
 
 
+def _df_serialize(x):
+    return jsonable_encoder(
+        {
+            'idx_type': 'RangeIndex' if isinstance(x.index, pandas.RangeIndex) else 'Array',
+            'df': _tuple_key_serialize(x.to_dict(orient='dict')),
+        }
+    )
+
+
 def _df_deserialize(x):  # pylint: disable=inconsistent-return-statements
     """
     A function to deserialize data into a DataFrame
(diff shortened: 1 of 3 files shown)


tamara-causalens (commented):
Hey @AidaZheng are there any other types we should be careful of?

AidaZheng:
> Hey @AidaZheng are there any other types we should be careful of?

I would say it is better to use `DataVariable` instead of `Variable`. When the df is serialised to a dict, some info is missing, like index type. It will be a great effort to find out all the cases and add all the missing info by ourselves.

I think for most of the case, the index should be correct when it is a array type data. This scenario is the index is a range type.

claudiawestby:
I can't use `DataVariables` because they can pass in multiple scenario results, either as a`List[pandas.DataFrame]` or if they want to name their results it also accepts `Dict[str, pandas.DataFrame]`. Both of these have to pass through this serialization. I guess the question is, and pardon my naivety in this area but  why can't the same serialization from a `DataVariable` be applied if it's recognized that a `pandas.DataFrame` is present in the object?

AidaZheng:
> I can't use `DataVariables` because they can pass in multiple scenario results, either as a`List[pandas.DataFrame]` or if they want to name their results it also accepts `Dict[str, pandas.DataFrame]`. Both of these have to pass through this serialization. I guess the question is, and pardon my naivety in this area but why can't the same serialization from a `DataVariable` be applied if it's recognized that a `pandas.DataFrame` is present in the object?

Only Variable's value is passing around between frontend to backend. Because it could be changed by user and backend does not know about it.  DataVariable's value is always stored in the backend and updated by action, so it is not passing around. Hope this make sense to you.
But, yep, we should figure out sth to fix your scenario. Is it possible to do:
```
df_1 = DataVariable(...),
df_2 = DataVariable(...),
@py_component
def show_result(df_1, df_2, dict):
     dict['key1'] = df_1
     dict['key2'] = df_2
     list = [df1, df2]
```
 Because the idea of Variable, it is some value can be changed in the UI, and then pass it to the backend. A DataFrame in theory couldn't be directly changed by UI, is updated by action.

krzysztof-causalens:
Yeah ideally you wouldn't have to store DataFrames in normal variables, as we can see here properly serializing them to JSON in a non-lossy way is very tricky; we can continue iterating on it for a while as we discover more and more edge cases but I'm positive the current solution won't be complete after this PR either. 

Perhaps we can take a step back and think about the use cases where DataVariables fall short. We have some improvements to them in the backlog, maybe it's worth thinking if that idea can be expanded to cover your use case as well and then channeling the resources to implementing that improvement rather than having to come back and tweak this serialization every couple weeks?

PR #287, opened 2024-05-07
Fix: DOS-1003 Address performance issues with useAction

sam-causalens:
## Motivation and Context
* Whilst not a massive performance issue in itself, useAction had a loading state that would trigger the component consuming the action to redraw whenever the action was called. The issue is that in a lot of cases useAction could be consumed near the top of component trees and in react contexts. This would lead to large tree re-resolutions whenever actions were run.

## Implementation Description
* To address this, we've moved the loading state of the action to be a Variable. A second hook "useActionIsLoading" has then be added so that any component that really cares about the loading state can still consume it easily.
* The added benefit here is now that `action.loading` can be consumed on the python side to conditionally alter other elements without needing py_components or hooking into the JS.

No

## How Has This Been Tested?
Covered by existing unit tests. Modifications made to account for the new api in a couple of places.

## PR Checklist:

A couple of performance traces of before and after this change. There's more work to do on this, but this is a good start.

Before:
Screenshot 2024-05-07 at 16 25 36

After (this one ran much more of the notebook in a similar time due to the speedup):
Screenshot 2024-05-07 at 16 25 57

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -14,7 +14,8 @@ from dara.components import Button
 def on_click(ctx: action.Ctx):
     # Do Something...
 
-Button('Click Me', onclick=on_click, disabled=on_click.loading)
+on_click_action = on_click()
+Button('Click Me', onclick=on_click_action, disabled=on_click_action.loading)
 ```
 
 -   On the JS Api side the `useAction` hook has been changed to only return the action function rather than a tuple of `[action_fn, isLoading]`. To retrieve the loading state a new hook `useActionIsLoading` now returns the isLoading state of the action as a piece of react state that will trigger redraws when its value changes.
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,23 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Updated AnnotatedAction to have a `loading` property that is automatically set to be a Variable[bool] instance tracking the loading state of the action. The example below shows how you can disable a button whilst the action it triggers is running.
+
+```python
+from dara.core import action
+from dara.components import Button
+
+@action
+def on_click(ctx: action.Ctx):
+    # Do Something...
+
+Button('Click Me', onclick=on_click, disabled=on_click.loading)
+```
+
+-   On the JS Api side the `useAction` hook has been changed to only return the action function rather than a tuple of `[action_fn, isLoading]`. To retrieve the loading state a new hook `useActionIsLoading` now returns the isLoading state of the action as a piece of react state that will trigger redraws when its value changes.
+
 ## 1.8.6
 
 -   Fixed an issue where `Chat` button looked off centre on different base font size apps.
@@ -19,7 +36,7 @@ title: Changelog
 
 -   `CausalGraphViewer` now only recalculates its layout on resize of the graph window if the graph is not in focus.
 -   Fixed an issue where `Select(..., multiselect=True)` would cause excessive rerenders and degrade performance when
-then number of items in the select was large.
+    then number of items in the select was large.
 
 ## 1.8.2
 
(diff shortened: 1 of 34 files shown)


packages/dara-components/js/common/button/button.tsx
```diff
@@ -69,7 +70,8 @@ const StyledButton = injectCss(styled(UiButton)<StyledButtonProps>`
  */
 function Button(props: ButtonProps): JSX.Element {
     const [style, css] = useComponentStyles(props);
-    const [onClick, loading] = useAction(props.onclick);
+    const onClick = useAction(props.onclick);
+    const loading = useActionIsLoading(props.onclick.loading);
```
krzysztof-causalens, line 74: Technically `props.onclick` can be `None` I believe, this should have a null check 
```suggestion
    const loading = useActionIsLoading(props.onclick?.loading);
```

packages/dara-components/changelog.md
```diff
@@ -2,6 +2,23 @@
...
+
+```python
+from dara.core import action
+from dara.components import Button
+
+@action
+def on_click(ctx: action.Ctx):
+    # Do Something...
+
+Button('Click Me', onclick=on_click, disabled=on_click.loading)
```
krzysztof-causalens, line 17: Shouldn't it be this way:

```suggestion
@action
def on_click(ctx: action.Ctx):
    # Do Something...
    
action_instance = on_click()

Button('Click Me', onclick=action_instance, disabled=action_instance.loading)
```

Since you can invoke the same action creator multiple times, they get separate loading states right?

sam-causalens: Ah, yeah fair point. I wasn't sure at what point it became an annotated action. Will update this.

packages/dara-components/js/common/button/button.tsx
sam-causalens: It shouldn't even have `.loading` here 🤦‍♂️. The null check is in the hook...

packages/dara-core/dara/core/base_definitions.py
```diff
@@ -423,6 +423,17 @@ def my_action(ctx: action.Ctx, ...):
...
+    loading: 'Variable'
+    """Loading Variable instance"""
+
+    def __init__(self, **data):
+        # Resolve the circular dependency to add a loading Variable to the model upon creation
+        # pylint: disable-next=import-error, import-outside-toplevel
+        from dara.core.interactivity.plain_variable import Variable
+
+        self.update_forward_refs(Variable=Variable)
+        super().__init__(**data, loading=Variable(False))
```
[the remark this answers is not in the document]
sam-causalens: Yep, took a while to figure out how to manage the import loop here 😬

packages/dara-core/tests/js/use-action.spec.tsx
```diff
@@ -938,10 +946,13 @@ describe('useAction', () => {
         const wsClient = new MockWebSocketClient('uid');
 
         const { result } = renderHook(
-            () =>
-                useAction(annotated, {
+            (): [(input?: any) => Promise<void>, boolean] => {
```
patricia-causalens, line 949: Wasn't this type changed to just have Promise<void>?

patricia-causalens (approved):
LGTM, just one small question

packages/dara-core/tests/js/use-action.spec.tsx
sam-causalens: Yes, but in this test hook I combine both the useAction and useActionIsLoading hooks into a single return so I can test both at the same time, hence the typing here.

patricia-causalens: Oh oops got the deleted part and added part mixed up thought you were typing useAction, thank you for explaining!

patricia-causalens (approved):

PR #290, opened 2024-05-09
Fix: DOS-1010 Fix the websocket reconnection so this.socket is correct

sam-causalens:
## Motivation and Context
* Found more bugs with the reconnection logic that reconnected the streams properly, but no messages could be sent back which broke most things.

## Implementation Description
* Previous change was refactored late on to have initialize return the new socket rather than setting it. This was not properly applied in the reconnect logic so we ended up with the stale socket on the class so no messages could be sent.

No

## How Has This Been Tested?
Unit test extended to cover it

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,24 +2,6 @@
 title: Changelog
 ---
 
-## NEXT
-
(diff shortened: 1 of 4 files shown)


krzysztof-causalens (approved):
Nice catch! LGTM

tamara-causalens (approved):

patricia-causalens (approved):

PR #293, opened 2024-05-13
Fix: DO-2977 preserve app url after login

krzysztof-causalens:
## Motivation and Context

This PR fixes an issue where the 'referrer' feature which was meant to redirect the user back to the original page after an auth error and a login was not working consistently.
This will be followed by updates to the internal SSO login system.

There were a couple of issues spotted:
- The backend auth endpoint had `HTTPBearer` FastAPI dependency with default arguments, which included `auto_error=True`; this causes it to raise its own error in case of missing token rather than let us handle it. Our client-side logic expects specific defined auth error messages which meant the referrer logic did not run for missing tokens. Now malformed auth headers (wrong scheme, missing token) explicitly return a 400 error
- The two main redirection parts in Dara, `handleAuthErrors` which runs on all API requests, and `<Redirect to="/login">` component rendered when a token is missing, did not explicitly pass through the referrer. These now correctly put `pathname` + `search` parts of the location in the `referrer` search param, URI encoded
- Default and Basic login pages now correctly URI decode the referrer and use it when navigating back after a successful login

## How Has This Been Tested?

Locally, see demo below
Updated backend unit test

## PR Checklist:

Demo

https://github.com/causalens/dara/assets/87647189/49a63d07-5c6a-4d32-9117-051323cd1bb3

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fixed an issue where desired pathname + search params were not retained when redirected to the login page and then back to the original page
+
 ## 1.9.0
 
 -   Further fix for the websocket reconnection logic that meant the socket class wasn't updated after re-initialization which meant that message sends were routed to the old socket rather than the new one.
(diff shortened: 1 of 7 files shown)


patricia-causalens (approved):
LGTM

PR #295, opened 2024-05-14
Fix: DO-2768 checkbox group and radio group docs

krzysztof-causalens:
## Motivation and Context

It's been reported that CheckboxGroup does not work properly in a following scenario:

```python
from dara.components.common import CheckboxGroup
from dara.components import Item

item1 = Item(value=1, label='foo')
item2 = Item(value=2, label='bar')

CheckboxGroup(
    items=[item1, item2],
    value=Variable([item1]),
)
```

It turns out this example works if we change the following:

```
CheckboxGroup(
    items=[item1, item2],
-    value=Variable([item1]),
+    value=Variable([1]),
)
```

This is a matter of a missing example in the docstring. The behaviour is consistent with other Select and Radio components - the `value` prop keeps track of the `value` selected, not the entire item. 

This PR fixes the CheckboxGroup and RadioGroup docs to correctly show this behaviour in examples.

## How Has This Been Tested?

Tested the behaviour locally

## PR Checklist:

### DIFF
--- a/packages/dara-components/dara/components/common/checkbox_group.py
+++ b/packages/dara-components/dara/components/common/checkbox_group.py
@@ -50,7 +50,7 @@ class CheckboxGroup(FormComponent):
 
     ```python
     from dara.core import Variable
-    from dara.components.common import CheckboxGroup
+    from dara.components.common import CheckboxGroup, Item
 
     CheckboxGroup(
         items=[Item(label='first',value=1), Item(label='second',value=2)],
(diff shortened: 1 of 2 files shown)


tamara-causalens (approved):

packages/dara-components/dara/components/common/checkbox_group.py
```diff
@@ -46,6 +46,18 @@ class CheckboxGroup(FormComponent):
...
     ```
 
+    CheckboxGroup with items with custom labels and values:
+
+    ```python
+    from dara.core import Variable
+    from dara.components.common import CheckboxGroup
+
+    CheckboxGroup(
+        items=[Item(label='first',value=1), Item(label='second',value=2)],
```
patricia-causalens, line 56: Missing Item import in the example

patricia-causalens (approved):
LGTM, just one small import missing

PR #297, opened 2024-05-15
Fix: DO-2997 card hiding some elements in internal product

patricia-causalens:
## Motivation and Context

When testing in internal product an example where in an app the content of the card showed but in the product it wsa hidden since it had essentially a hug component around it.

## Implementation Description

Since card children did not have flex-grow set, it meant that in this specific scenario where the children were scrollable stacks that they ended up with height being calculated as 0. The suggested fix is to default flex-grow to 1 instead.

None

## How Has This Been Tested?

Locally on the internal product that the fix worked and rendered components app to check if the fix affected anything else.

## PR Checklist:

### DIFF
--- a/borg.toml
+++ b/borg.toml
@@ -1,1 +1,1 @@
-version = "1.8.6"
+version = "1.9.1"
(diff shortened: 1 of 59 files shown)


patricia-causalens:
> Hey @patricia-causalens, could you please add a screenshot of before and after if possible?

Sure, here is the before:
Screenshot 2024-05-15 at 22 04 34
and after:
Screenshot 2024-05-15 at 22 06 01

krzysztof-causalens (approved):

claudiawestby (approved):
thanks for the fix!

PR #298, opened 2024-05-16
Fix: DO-3014 multiselect value outside of items

krzysztof-causalens:
## Motivation and Context

Just like `Select`, it was intended that select with `multiselect=True` should display its `value` as a fallback even when the items are not in the `items` list.

There was a minor bug which caused the effect to wipe the selected items after the initial render.

## How Has This Been Tested?

Locally

```py
config.add_page('test', Select(multiselect=True, items=Variable([]), value=Variable(['one'])))
```

https://github.com/causalens/dara/assets/87647189/72971883-0f52-4c56-9eee-f3992bb56997

## PR Checklist:

### DIFF
--- a/packages/dara-components/js/common/select/select.tsx
+++ b/packages/dara-components/js/common/select/select.tsx
@@ -159,7 +159,7 @@ function Select(props: SelectProps): JSX.Element {
     const itemArray = formattedItems as Array<Item>;
 
     if (props.multiselect) {
-        const explicitValues = useMemo(() => isArray(value) ? value.map(toItem) : value, [value]);
+        const explicitValues = useMemo(() => (isArray(value) ? value.map(toItem) : value), [value]);
         const foundItems = getMultiselectItems(value, itemArray);
         const [selectedItems, setSelectedItems] = useState(isEmpty(foundItems) ? explicitValues : foundItems);
         const onSelect = useCallback(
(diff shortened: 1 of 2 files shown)


packages/dara-components/js/common/select/select.tsx
```diff
@@ -175,7 +176,8 @@ function Select(props: SelectProps): JSX.Element {
         // it fail to update the value sometimes. By keeping a local copy of the value and updating it like this we fix
         // the race condition and respect the main value if it is updated elsewhere.
         useEffect(() => {
-            setSelectedItems(getMultiselectItems(value, itemArray));
+            const found = getMultiselectItems(value, itemArray);
+            setSelectedItems(isEmpty(found) ? explicitValues : found);
```
patricia-causalens, line 180: I think in the select version we check if explicitValues is undefined as well if so we set it to null, I can't remember why but have you tried to see if that is ok here?

patricia-causalens (approved):
LGTM, just one question

tamara-causalens (approved):

packages/dara-components/js/common/select/select.tsx
krzysztof-causalens: I think it was to differentiate undefined/null for controlled component purposes. I'll do the same here to be on the safe side

PR #300, opened 2024-05-17
Improvement: DO-3021 allow `None` children, component serialization improvements

krzysztof-causalens:
## PR Checklist:


packages/dara-components/js/common/stack/stack.tsx
```diff
@@ -16,7 +16,7 @@ import styled from '@darajs/styled-components';
 // The props are actually used deep within the useComponentStyles etc hooks
 /* eslint-disable react/no-unused-prop-types */
 interface StackProps extends LayoutComponentProps {
-    children: Array<ComponentInstance>;
+    children: Array<ComponentInstance | null>;
```
patricia-causalens, line 19: Do we need to update this in more places for null? Not sure I understand why it was only updated on Stack

krzysztof-causalens: I should actually revert it, I started implementing it here but moved it to base component instead.
DynamicComponent should make sure it's not actually null, good catch

packages/dara-components/dara/components/common/stack.py
```diff
@@ -137,5 +137,5 @@ class Stack(LayoutComponent):
     scroll: bool = False
 
     # Dummy init that just passes through arguments to superclass, fixes Pylance complaining about types
-    def __init__(self, *args: ComponentInstance, **kwargs):
+    def __init__(self, *args: Union[ComponentInstance, None], **kwargs):
```
patricia-causalens, line 140: Do we need to add this union to other components as well? e.g. card, grid and others that accept ComponentInstance?

krzysztof-causalens: Yep, I missed those two, thanks

patricia-causalens (approved):
LGTM!

PR #311, opened 2024-06-10
Fix: DO-3050 DerivedDataVariable leaking PendingValue to userspace

krzysztof-causalens:
## Motivation and Context

It's been reported that in some cases DerivedDataVariable (DDV) can be invoked with a `PendingValue`, an internal Dara store value which represents a started computation. Upon investigation, this could happen in a scenario such as:
- a DDV computation is invoked and finished, resolving to a cache key to query data from
- another computation is invoked for the same set of inputs
- while the second computation is ongoing, a query for the data is made against the cache key

This caused the data query to retrieve a `PendingValue` from the store which was inserted there by the second computation. 

The fix was to handle that case and await the computation resolution.

## Implementation Description

As part of the investigation I spotted another issue where a cache eviction could get 'inbetween' the derived and query parts of the DDV, causing the DDV to resolve to None and error. I've added a `pin` to the derived part of the query which should ensure the cache entry won't be evicted at the very least until the data is read back.

There can still be rare edge cases where multiple subsequent queries and recomputations can cause an eviction of a data point which is being queried. They should be minor enough that I decided not to look into them too deeply for now. A potential solution would involve tying in the Data and Derived parts of the caching in some way. Perhaps a rework of the DataVariable APIs (which has been in plans, longer term) can solve it in a different way.

## How Has This Been Tested?

Added two unit tests, recreating the scenarios which broke beforehand.

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fixed an issue where in some cases a `DerivedDataVariable` would be invoked with an internal `PendingValue`
+
 ## 1.9.1
 
 -   Fixed an issue where desired pathname + search params were not retained when redirected to the login page and then back to the original page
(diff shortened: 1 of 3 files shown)


patricia-causalens (approved):
LGTM

tamara-causalens (approved):
Nice find!

PR #313, opened 2024-06-11
Docs: DO-3096: added FAQ to card issue found

patricia-causalens:
## Motivation and Context

Claudia found an issue where in her component the content of the card would overflow its container. A simplified version can be seen below where on the left we see that the blue overflows the coral card:
Screenshot 2024-06-11 at 11 27 26

## Implementation Description

I looked into ways we could address it, for example we could set the `margin-bottom` of the `Card` to `auto`, which does seem to allow it to expand in this scenario. However in other scenarios that would cause a spacing to appear below the child components instead of filling all the space. We could also set the `height` to `fit-content` instead, but this can result in the same issues where the `Card` doesn't fill the available space if the content is small.

I am not entirely sure why wrapping it in a `Stack` works, I have tried adding the same css properties to `Card` and couldn't spot exactly what the difference is. As far as logic goes `Card` behaves as I would expect from these components in this scenario.

One thing that I noticed as a difference is that Stack has a chain of 100% height all the way to the root of the page, whereas for `Card` this chain is broken, indeed if we set height to 100% for the `CardDiv` we see that `Stack` then also causes it to overflow the `Card` container! Some obscure CSS magic seems to be at work here here.

Given this seems like a very specific and rare occurrence and it is quite easy for the user to wrap a `Stack` around the `Card`, I have added an FAQ entry for now. We can revisit this and explore deeper if it comes back up again.

## PR Checklist:


docs/faq.md
```diff
@@ -338,4 +338,10 @@ import plotly.graph_objects as go
 fig = go.Figure(layout=dict(template='plotly'))
 ```
 
+</details>
+<details>
+<summary><h3 style={{margin: '0px 0px 0px 0px'}}>My `Card` is not fitting its content, what do I do?</h3></summary>
```
tamara-causalens, line 343: can you maybe add an image as an example?

Roman-Kornev: I would also suggest adding the code fix example

Roman-Kornev (approved):

docs/faq.md
patricia-causalens: Good points, have added both

tamara-causalens (approved):

PR #316, opened 2024-06-12
Fix: DO-3052 create from derived variable is not resolved correctly as a py component dependency

patricia-causalens:
## Motivation and Context

An issue was found where if one was creating a `Variable` from a `DerivedVariable` that in some cases the whole `DerivedVariable` was returned instead of its value.

## Implementation Description

Upon closer inspection this turned out to be the case when the `Variable` had `persist_value` set to true or when it was embedded as is the case in parts of the internal product. 

This came to where the effect was set for the Variable atom, and the read value was set to `variable.default` which in this case is the DerivedVariable. I have updated to instead detect if it is derived and if so get its value.

None

## How Has This Been Tested?

Locally on a test app, and by adding a test for this scenario

## PR Checklist:


patricia-causalens:
I am fixing the cypress test otherwise good for review

krzysztof-causalens (approved):
LGTM, nice job. Not a pretty fix but I'm not sure there is a nice way in Recoil to just keep the default behaviour so I guess we have to carry around the extra state into effects.

PR #319, opened 2024-06-19
Fix: DO-3161: Fix infinite loop when a nested selector is used in Select

Roman-Kornev:
## Motivation and Context

In certain conditions when the Select `value` is a nested dictionary selecting a new value from the dropdown causes an infinite loop. The reason seems to be related to an edge case in an interaction between Recoil and Downshift causing the `onSelect` to retain a stale value after forcing a rerender, which then tries to set the value back to its current state.

```py
from dara.components import Select
from dara.core import Variable, py_component, DerivedVariable

dict_var = Variable({x: None for _ in range(10)})
other_var = Variable(0)
    
derived_var = DerivedVariable(
    lambda x, y: True, variables=[dict_var, other_var],
    deps=[other_var]
)

@py_component
def MyComponent(_):
    return Select(items=['a', 'b', 'c'], value=dict_var.get(6))

MyComponent(derived_var)
```
It's not always reproducible, on average about 20% of page loads will ever have this issue, and not every select click will trigger it. The timing of selecting an option also seems to have an impact but it's still unclear how to reliably reproduce it.

## Implementation Description

The fix involves 2 things:
- Using a ref to access the current value in `onSelect` and only setting the value if it's different.
- Returning an immediate `useRecoilState` hook instead of a deferred loadable in `useVariable`.

As part of this, the previous hack of retaining a local copy of the selected state in the Select component is no longer necessary.

## How Has This Been Tested?

Manually

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/shared/interactivity/use-variable.tsx
+++ b/packages/dara-core/js/shared/interactivity/use-variable.tsx
@@ -82,7 +82,11 @@ export function useVariable<T>(variable: Variable<T> | T): [value: T, update: Di
 
     const recoilState = getOrRegisterPlainVariable(variable, WsClient, taskContext, extras);
     if (!isDerivedVariable(variable.default)) {
-        return useRecoilState(recoilState);
+        const [value, setValue] = useRecoilState(recoilState);
+        useEffect(() => {
+            bus.publish('PLAIN_VARIABLE_LOADED', { variable, value });
+        }, [value]);
+        return [value, setValue];
     }
     const [loadable, setLoadable] = useRecoilStateLoadable(recoilState);
     useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/dara-core/js/shared/interactivity/use-variable.tsx
```diff
@@ -81,16 +81,18 @@ export function useVariable<T>(variable: Variable<T> | T): [value: T, update: Di
     }
 
     const recoilState = getOrRegisterPlainVariable(variable, WsClient, taskContext, extras);
+    if (!isDerivedVariable(variable.default)) {
```
krzysztof-causalens, line 84: We probably still need the effect  (or a similar one) below to run in this branch as well, since our internal products rely on the bus event below being fired for all variable changes. E.g. you could add

```tsx
const [value, setValue] = useRecoilState(recoilState);
useEffect(() => {
 bus.publish(...value...)
}, [value]);
return [value, setValue];
```

krzysztof-causalens (changes requested):

krzysztof-causalens (approved):

PR #320, opened 2024-06-19
Improvement: DOS-1049 - added id and for the ComponentInstance

momustafa-causalens:
## Motivation and Context
Would like a way to tie components to one another using the `id` and `for` pattern.

## Implementation Description
Added two optional properties to `ComponentInstance` following the [PEP 8 Style Guide](https://peps.python.org/pep-0008/#function-and-method-arguments)'s recommendation on properties that clash with reserved keywords.

This is not expected or required to show up in the DOM at this stage.

## How to use
To any component that inherits from `ComponentInstance` you can now add an `id_` and `for_` property
which will then persist in the serialized json form of that component, to be used as the user pleases.
This can help identify components with human-readable names in the serialized trees.

No

## How Has This Been Tested?
Manually

## PR Checklist:


krzysztof-causalens (approved):

Roman-Kornev (approved):
Can you add to the PR description and to the docs some info on how we're supposed to use these?

patricia-causalens (approved):
LGTM

PR #323, opened 2024-06-27
Fix: DO-3214 input not firing `onchange` action

krzysztof-causalens:
## Motivation and Context

Issue reported in an internal product that state does not behave as expected. After investigating the issue, I noticed the `onchange` action did not fire at all for an Input component. 

Looking for similar code in other components, the same logic applied in `Textarea` so I applied the fix there too. 

The reason why this regressed now is I think due to #319. There we changed simple Variables to no longer go through a deferred layer which made the state update a tick later (in an effect). This made the current, buggy code work and the recent change unveiled the bug.

## Implementation Description

Digging deeper I spotted that the action was being scheduled as a debounced call but then immediately cancelled by an effect.
The fix was to simply not cancel scheduled debounced action calls to prevent these issues from happening. I left the other debounce cancels in place as those are needed to make sure outside state changes are not overriden by accident.

## How Has This Been Tested?

Manually in a test app and in internal product

Test code:

```python
var = Variable()
target = Variable()

@action
async def test(ctx):
    await ctx.update(target, ctx.input)

Stack(
   Input(value=var, onchange=test()),
   Textarea(value=var, onchange=test()),
   Text('Value: '),
   Text(var),
   Text('Target: '),
   Text(target),
)
```

https://github.com/causalens/dara/assets/87647189/da159a4f-615b-42da-bb4f-95cfa51fcd9b

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Fixed an issue where `onchange` would not fire for `Input` and `Textarea` components in some cases
+
 ## 1.10.0
 
 -   Added support for grouping layout in `FcoseLayout` and `SpringLayout`. It allows for nodes to be placed on groups/clusters which can be collapsed/expanded in order to simplify large graphs to the user whilst still allowing them to see the full picture.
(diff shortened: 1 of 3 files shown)


patricia-causalens (approved):
Good find! LGTM

Roman-Kornev (approved):
Great find, thanks!

PR #326, opened 2024-06-28
Fix: DO-3172 numeric input none value is inconsistent

patricia-causalens:
## Motivation and Context

Issue was found when emptying a numeric input. Its value as stored in the variable was `None`, whereas in an action `ctx.input` returned `NaN`. 

## Implementation Description

We convert the value based on whether the input is numeric, however we were not passing this new value to the action

None

## How Has This Been Tested?

Locally

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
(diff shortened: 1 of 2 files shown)


krzysztof-causalens (approved):
Nice find!

PR #327, opened 2024-06-28
Fix: DO-3224 EventBus events not firing for actions updating variables

krzysztof-causalens:
## Motivation and Context

There are cases where a Dara variable is not directly used by a `useVariable` hook, it can still 'live' in client-side state and be updated by actions. E.g.

```python
from dara.components import Input

sync_to = Variable()

Input(onchange=sync_to.sync())
```

In systems relying on the event bus events to track the current state of the variable, in the scenario above there was never an event fired about the variable state changing since only `useVariable` used to fire events.

This PR adds explicit events about updated variables to the variable-mutating actions.

## Implementation Description

Added event bus to the context injected into actions, grabbing from nearest capturer with `useEventBus`. Actions then simply publish events when they mutate variables

## How Has This Been Tested?

Unit tests added
Tested code in internal product

## PR Checklist:


patricia-causalens (approved):
LGTM!

PR #330, opened 2024-07-03
Build:1.10.5 + npmrc fix

krzysztof-causalens:
## Motivation and Context

Releasing with a small fix to Dara-core build scripts. In scenarios where we want a more customized .npmrc we can create it ourselves, but the script overwrites it with an empty .npmrc currently. This fix changes is so the empty .npmrc is written only if a custom registry is passed to the build script, otherwise it's left as is.

## How Has This Been Tested?

Locally

## PR Checklist:


patricia-causalens (approved):
LGTM

Roman-Kornev (approved):

PR #339, opened 2024-07-12
Improvement: DO-2951: Upgrade pixijs to v8

Roman-Kornev:
## Motivation and Context

Upgrades pixijs to v8.

Added an experimental `enable-webgpu` localstorage flag to preview webgpu renderer. From initial testing performance should be roughly the same.

Other changes:
- Dropped `SmoothGraphics` as it's not compatible with v8. Hopefully in the future we can bring it back.
- Dropped `pixi-essentials/cull`. The builtin culling plugin from v8 seems to be working just fine for our usecase.
- Capped `pixi-viewport` to 5.0.3 as their latest release on NPM is not the same as the one on Github, reverting some changes.

## How Has This Been Tested?

Locally on MacOS and Windows.

## PR Checklist:


packages/ui-causal-graph-editor/src/shared/rendering/background.tsx
```diff
@@ -26,7 +26,7 @@ import { colorToPixi, createKey } from './utils';
 const DOT_DISTANCE = 20;
 const SCALE = 0.5;
 
-export class Background extends PIXI.utils.EventEmitter<'click'> {
+export class Background extends EventEmitter<'click'> {
```
krzysztof-causalens, line 29: This can just be `PIXI.EventEmitter` without the extra import

packages/ui-causal-graph-editor/src/shared/rendering/node/node-object.tsx
```diff
@@ -239,14 +240,14 @@ export class NodeObject extends PIXI.utils.EventEmitter<(typeof MOUSE_EVENTS)[nu
                 const textStyle = getTextStyle(nodeStyle.label_size);
                 const trimmedText = trimToFit(nodeStyle.label, maxSize, textStyle);
 
-                const txt = new PIXI.Text(trimmedText, textStyle);
-                txt.resolution *= 2;
+                const txt = new PIXI.Text({ text: trimmedText, style: textStyle });
+                txt.resolution = 2;
```
krzysztof-causalens, line 244: Was it always 1 before? I remember in some cases some resolution numbers were different depending on the display (retina vs other), there might have been a reason we were multiplying here

krzysztof-causalens (approved):
Looks good. Have you had a look at the culling performance and whether it differs from our previous manual culling?

patricia-causalens (commented):
Nice work, was just playing with it locally and it seems that fcose grouping example takes significantly longer to collapse first time with these changes, do you know what could have caused that? It likely will need other performance improvements there later as well, but was curious.

Roman-Kornev:
@patricia-causalens Thanks for checking it out. The perfrormance issue you were seeing is not actually the collapsing process being slow, it's that the graph is fully invisible until you move the viewport, which makes it seem like it's slow. 

Turns out the new culler is sometimes behaving strangely when we delete and recreate lots of nodes at once. The old culler is not ported to v8, but I did manage to find a workaround by setting the dirty flag on the viewport. Should work fine for now. We can look into implementing some more fine-grained culling approaches, as right now it just culls everything marked as cullable.

patricia-causalens (approved):
Thanks for looking into it, looks better now! LGTM once CI passes!

packages/ui-causal-graph-editor/src/shared/rendering/engine.tsx
```diff
@@ -545,6 +545,11 @@ export class Engine extends PIXI.EventEmitter<EngineEvents> {
             });
 
             this.requestRender();
+
+            // Force a culling update all nodes and edges are created
```
patricia-causalens, line 549: could we maybe just add why we need this for now if it is a workaround

PR #340, opened 2024-07-16
Security: DO-3367 set minimum version for certifi

patricia-causalens:
## Motivation and Context

Certifi library has a CVE against it: https://github.com/advisories/GHSA-248v-346w-9cwc

Added a minimum version for it, set to the patched version number as per the report above - `2024.7.4` 

## PR Checklist:


zack-causalens (approved):

pyproject.toml
```diff
@@ -12,6 +12,7 @@ anyio = ">=4.0.0"
 async-asgi-testclient = ">=1.4.11,<2.0.0"
 bokeh = ">=3.1.0,<3.2.0"
 cai-causal-graph = ">=0.3.6"
+certifi = ">=2024.7.4"
```
krzysztof-causalens, line 15: That's just the top-level pyproject which will be regenerated by anthology. The minimum version should be set in the packages themselves.

krzysztof-causalens (changes requested):

pyproject.toml
patricia-causalens: Oops, good catch fixed it

packages/dara-core/pyproject.toml
```diff
@@ -50,7 +51,7 @@ openpyxl = "*"
 packaging = "^23.1"
 pandas = ">=1.1.0, <3.0.0"
 prometheus-client = "^0.14.1"
-pyarrow = "*"
+pyarrow = "<17.0.0"
```
krzysztof-causalens, line 54: What's the reason to restrict pyarrow here?

patricia-causalens: Was just testing, the version had just been released and was not being picked up yet so was just checking the issue was not elsewhere. Seems to be picking it up now though so have removed it

krzysztof-causalens (approved):

PR #341, opened 2024-07-22
Fix: DO-3469 vfile version

krzysztof-causalens:
## Motivation and Context

Fixing vfile version to a known working version as the new patch release is breaking production rollup builds.

Relevant issue: https://github.com/vfile/vfile/issues/90

Usual error looks like this:

```
[vite]: Rollup failed to resolve import "#minpath" from "node_modules/vfile/lib/index.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
error during build:
Error: [vite]: Rollup failed to resolve import "#minpath" from "node_modules/vfile/lib/index.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
```

Confirmed in an internal product that locking to the version before import maps were introduced fixes the issue. Could be because of our relatively old Vite/Rollup version but we don't have the bandwidth to upgrade right now so locking the dependency instead.

## PR Checklist:


tamara-causalens (approved):

patricia-causalens (approved):
LGTM

PR #342, opened 2024-07-22
Feat: DO-3423 update code viewer to allow copy to clipboard

patricia-causalens:
## Motivation and Context

We would like to use our code component within chat component, so that code snippets will look better than the default with markdown and allow it so that code can be copied with a button.

## Implementation Description

To achieve this I have moved the Code component to `ui-components` package, this way `Chat` component will be able to easily import it for the next step.

I have added a copy to clipboard button, and since this will be in the future part of the chat, I have also added a top bar to the component which shows the language. This also helps as a way of preventing the clipboard button overlaying the code text in the narrow space within a chat. 

While looking into it I have also updated the default theme for both light and dark modes. 

`prism-react-renderer` was added to `ui-components`

## How Has This Been Tested?

Locally on storybook and on a dara test app

## PR Checklist:

code viewer

Screenshot 2024-07-22 at 16 44 37


packages/ui-components/src/code-viewer/code-viewer.tsx
```diff
@@ -0,0 +1,156 @@
...
+            if (props.codeTheme === CodeComponentThemes.LIGHT) {
+                return nightOwlLight;
+            }
+            return vsDark;
+        }
+        if (themeCtx.themeType === 'dark') {
+            return vsDark;
+        }
+        return nightOwlLight;
+    }
```
krzysztof-causalens, line 114: This could be a memo instead of a getter function

pnpm-lock.yaml
```diff
@@ -270,6 +270,24 @@ importers:
       vite: 3.1.8
       whatwg-fetch: 3.6.20
 
+  packages/dara-demo/dist:
```
krzysztof-causalens, line 273: Looks like you had a demo package in your local repo, could you clean up the lockfile? (recreate with a clean repo)

krzysztof-causalens (approved):

pnpm-lock.yaml
patricia-causalens: Oops forgot to commit the lockfile after removing it, good catch!

tamara-causalens (approved):

PR #344, opened 2024-07-22
Improvement: DO-3453 make chat component more flexible and UI updates

patricia-causalens:
## Motivation and Context

Update chat component in order to make it more flexible and with latest design

## Implementation Description

- Updated chat component so that it has a non "popup" version that can be used to embed in other places. 
- Updated so that messages sent by the user now show in a different color.
- Updated `Textarea` component so that as you type the are grows for chat component usage
- Added aria roles etc to improve the component's accessibility

None

## How Has This Been Tested?

Locally on storybook and a test app

## PR Checklist:

chat


packages/ui-components/src/chat/chat.tsx
```diff
@@ -99,6 +98,12 @@ export interface ChatProps extends InteractiveComponentProps<Message[]> {
     onUpdate?: (value: Message[]) => void | Promise<void>;
     /** The user who is currently active in the chat */
     activeUser: UserData;
+    /** The title to appear at the top of the chat */
+    chatTitle?: string;
+    /** The placeholder text for the reply field */
+    placeholder?: string;
+    /** Whether the chat is in a popup and should be styled as such */
+    isPopup?: boolean;
```
krzysztof-causalens, line 106: Would it be more composable instead if you exposed effectively this component without the wrapper (so 'ChatBody', ReplyWrapper + the logic, intended to be rendered directly) as a 'Chat' component and then a 'ChatPopup' which would wrap the first component? Or are the popup visuals partially in dara-components? 
Feels like it could be a nicer pattern than having to do conditional styling, this way the place where we use it can choose to use its own wrapper if need be rather than be stuck with the two provided options. 

If it's urgent I'm fine with the current approach but I think it could maybe be improved

krzysztof-causalens (approved):

tamara-causalens (approved):
Very nice!

PR #345, opened 2024-07-23
Improvement: DO-3034 support seamless handling of refresh tokens

krzysztof-causalens:
## Motivation and Context

This PR introduces handling of refresh tokens to the framework core. 

When any request fails with a 401 or 403 status code, we initiate a token refresh process. 
If the current auth config supports token refresh, it can choose to use the previous expired token and attached `dara_refresh_token` cookie to produce a new session token.
If successful, the token is replaced in a new global store and the request is retried.
If unsuccessful, the original failed response is returned to the caller.

### Client-side

Session token has been moved out of a React context into a singleton global store. This gives us more control over refreshing tokens. For example, once a token refresh is already initiated, other attempts to retrieve the token will block and wait for the in-flight refresh to go through. 

The above is achieved utilizing a simple store with a locking mechanism, where in-flight refreshes are represented by promises.

Auth internals like the `useSessionToken` hook have been updated to pull/subscribe to the store instead of the context. 

### Backend

The `refresh-token` endpoint is now included by default in dara core. The base config provides an implementation of a `refresh_token` method which raises a 400 - refresh not supported - error. 

An auth config can be configured to support token refresh by:
- implement the aforementioned method to sign a new token, reusing the previous session_id for continuity
- add some mechanism to set-cookie with the `dara_refresh_token` used, e.g. via custom `components_config` and endpoints such as `sso-callback` for SSO auth

The above is utilized in the internal SSO auth package.  

## How Has This Been Tested?

Unit tests added
Manually tested with internal SSO auth implementation

## PR Checklist:


packages/dara-core/js/shared/global-state-store.tsx
```diff
@@ -0,0 +1,134 @@
...
+        let result: TState[TKey];
+
+        try {
+            result = await fn();
+            this.#state[key] = result;
+
+            // success - notify subscribers, unlock pending promises and set the value
+            this.#notify(key, result);
+            unlock(result);
+            this.setValue(key, result);
```
Roman-Kornev, line 92: Isn't this `setValue` doing the same thing as 3 lines above?

Roman-Kornev (approved):

packages/dara-core/js/shared/global-state-store.tsx
krzysztof-causalens: Good catch, yeah we'll notify twice here. I'll remove the explicit one 👍

packages/dara-components/changelog.md
```diff
@@ -2,7 +2,7 @@
 title: Changelog
 ---
 
-# NEXT
+## 1.14.0-alpha.0
```
patricia-causalens, line 5: We should be careful here to not forget to update to the proper version later

patricia-causalens (approved):
LGTM, just missing changelog entry

packages/dara-components/changelog.md
krzysztof-causalens: Good catch, I'll reset this to NEXT before merging

packages/dara-core/js/auth/auth-wrapper.tsx
```diff
@@ -1,16 +1,54 @@
...
+async function checkTokenExpiry(token: string): Promise<string | null> {
+    // check if token is expired
+    if (token) {
+        try {
+            const decoded = jwtDecode<{
+                exp: number;
+            }>(token);
+
+            // If it's about to expire within 5 minutes
+            if (decoded.exp * 1000 - Date.now() < 5 * 60 * 1000) {
```
sam-causalens, line 30: I think 5 minutes is too long. Our target time for the token length is only 10minutes, so this will lead it to refreshing every 5 minutes which seems excessive. I'd be okay with either not having this at all as I think it overcomplicates it for not much benefit (as everything will now block and wait for the refresh anyway?) or reducing this timer down to 10-15 seconds?

packages/dara-core/js/auth/auth-wrapper.tsx
```diff
@@ -48,38 +86,61 @@ function DynamicAuthComponent(props: { component: AuthComponent }): JSX.Element
...
 
-        setToken(newToken);
+    // set initial token from local storage as soon as we render
+    if (!isMounted.current) {
+        isMounted.current = true;
+        globalStore.setValue('sessionToken', getToken());
     }
 
+    // check token expiry every minute
+    useEffect(() => {
```
sam-causalens, line 112: What was the reasoning behind this interval? Why not just wait for a request to fail and then refresh the token?

packages/dara-core/js/shared/global-state-store.tsx
```diff
@@ -0,0 +1,133 @@
...
+     */
+    public async replaceValue<TKey extends keyof TState>(
+        key: TKey,
+        fn: () => Promise<TState[TKey]>
+    ): Promise<TState[TKey]> {
+        if (this.#locks[key]) {
+            return this.#locks[key] as Promise<TState[TKey]>;
+        }
+
+        let unlock: (res: TState[TKey]) => void;
```
sam-causalens, line 74: Can you add some comments in here to explain to someone later what is happening

sam-causalens (changes requested):
Looks Good Krys, left a couple of comments to think about.

packages/dara-core/js/auth/auth-wrapper.tsx
krzysztof-causalens: Sure, I agree the way things work now we can probably safely remove it and just let requests fail.

packages/dara-core/js/auth/auth-wrapper.tsx
krzysztof-causalens: To be fair I can't quite recall, I think the idea originated when we were targeting a longer refresh time so we were trying to refresh ahead of time. With the low token time now we can get away without it.

krzysztof-causalens:
- removed periodic token refresh, is not necessary with expected refresh token times and is redundant as we now gracefully handle failed requests by retrying them after refreshing
- removed `jwt-decode` dependency as we no longer have to decode the token on the client side to check it's expiry date

krzysztof-causalens:
I realized there was an issue where in-progress live WS connections would have stale ContextVar values, such as ID_TOKEN if one is used by an auth config. 

ContextVars can't be updated from the outside, so I've added a new memory object stream into the WS handler object:
- The WS handler logic now checks for pending token data updates in the receive stream and updates the context whenever one is found
- The refresh endpoint sends a token data update into the send stream

Added a test to verify the above behaviour. 

Tagging another alpha version - I'll test how that interacts with multi-tab tomorrow, I have a feeling we might run into issues when two tabs decide to refresh their tokens simultaneously

krzysztof-causalens:
Looking more into the cross-tab issues, we've decided to take a different approach.

1) The client-side `GlobalStore` used to store the session token is instead a wrapper around `localStorage`. This means the values are shared cross-tab, and subscriptions use `storage` events so one tab performing a refresh will automatically notify all active tabs.
2) To update live WS connections, the client sends a special `token_update` WS message for the backend to update its in-memory ContextVars.
3) To prevent multiple tabs from performing concurrent token refreshes, the calls to actual token refresh logic are wrapped with a short-lived deduping async cache. Simultaneous attempts to refresh the same token will be deduped and resolve to the same value (due to underlying locks), and subsequent ones within a short window (5s) will also return the same cached value.

packages/dara-core/dara/core/auth/utils.py
```diff
@@ -100,3 +102,120 @@ def get_user_data():
...
+
+
+token_refresh_cache = AsyncTokenRefreshCache(ttl_seconds=5)
+"""
+Shared token refresh cache instance
+"""
+
+
+async def cached_refresh_token(
+    func: Callable[[TokenData, str], Tuple[str, str]], old_token_data: TokenData, refresh_token: str
```
Roman-Kornev, line 189: `func` should probably be something like `do_refresh_token`

Roman-Kornev (approved):

sam-causalens (approved):
This LGTM, thanks for this Krys much appreciated.

PR #346, opened 2024-07-24
Feat: DO-3454 add loading state to chat component

patricia-causalens:
## Motivation and Context

Adds the ability to pass a custom component for the loading state of the chat, which when active blocks any further messages to be sent

## Implementation Description

Added a new prop to `Chat` to pass a component in order to show a "Loading" state which blocks user's messages.

None

## How Has This Been Tested?

Locally on storybook

## PR Checklist:

Screenshot 2024-07-24 at 10 31 18


tamara-causalens (approved):

packages/ui-components/src/chat/chat.tsx
```diff
@@ -217,6 +217,7 @@ function Chat(props: ChatProps): JSX.Element {
                         isEditable={didUserWriteMessage(message, props.activeUser)}
                     />
                 ))}
+                {props.loadingComponent}
```
krzysztof-causalens, line 220: Is the expectation that the caller is supposed to set this prop to null while the chat is not loading? Might be worth making that clear in the prop docstring.

krzysztof-causalens (approved):

packages/ui-components/src/chat/chat.tsx
patricia-causalens: Yes, that is what I was thinking, will update docstrings

PR #348, opened 2024-07-24
Feat: DO-3456 extract code from markdown and use code viewer in chat component

patricia-causalens:
## Motivation and Context

Improve the visuals of code blocks rendered within `Chat` component, and allow users to easily copy its content

## Implementation Description

~~Added a parser that breaks the message string into `Markdown` and `CodeViewer` components. This is only applied to the displayed message so that when editing users still have full control of the markdown string and can make changes if needed~~

Updated base markdown component to use `CodeViewer` for code blocks with a specified language.

## How Has This Been Tested?

Locally on storybook

## PR Checklist:

Screenshot 2024-07-24 at 21 20 57

### DIFF
--- a/packages/ui-components/src/chat/message.tsx
+++ b/packages/ui-components/src/chat/message.tsx
@@ -301,7 +301,7 @@ function MessageComponent(props: MessageProps): JSX.Element {
             )}
             {!editMode && (
                 <MessageBody>
-                    <Markdown markdown={localMessage.message} />
+                    <Markdown markdown={processText(localMessage.message)} />
                 </MessageBody>
             )}
         </MessageWrapper>
--- a/packages/ui-components/src/chat/message.tsx
+++ b/packages/ui-components/src/chat/message.tsx
@@ -16,14 +16,12 @@
  */
 import { format, parseISO } from 'date-fns';
 import { isEqual } from 'lodash';
-import { Language } from 'prism-react-renderer';
 import * as React from 'react';
 
 import styled, { useTheme } from '@darajs/styled-components';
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/chat/message.tsx
```diff
@@ -192,6 +194,55 @@ export function processText(text: string): string {
     return text.replace(/\n/g, '\n\n');
 }
 
+/**
+ * A function to parse markdown such that code blocks return CodeViewer component
+ * @param message
+ */
+function parseMarkdown(message: string): React.ReactNode[] {
```
krzysztof-causalens, line 201: I think we could simply utilize the tag->component map like in the example here https://github.com/remarkjs/react-markdown?tab=readme-ov-file#use-custom-components-syntax-highlight rather than relying on our own regex?

krzysztof-causalens:
Updated the approach slightly. Baked the code display into the base markdown display instead, if the code block has a specified language e.g. "```python...". 
If no language is specified or it's an inline code block e.g. `some code` then default display is used. 

Screenshot 2024-07-25 at 12 14 03

@tamara-causalens let me know if this behaviour looks okay.

krzysztof-causalens:
Dark mode doesn't look great in general yet but code blocks do switch to the opposite color which looks pretty good imo.

Screenshot 2024-07-25 at 12 23 11

tamara-causalens (approved):
Looks amazing! 😍

jsumiguin-causalens (approved):
LGTM, and I think it's fine for now to only allow code blocks with a language specified as AgentQA returns it anyways, but we might need to update this to do that sometime in the future.

PR #349, opened 2024-07-25
Improvement: DO-3471 - Improved some chat functionality

jsumiguin-causalens:
## Motivation and Context

Currently, the `localValue` for this component gets reset to `props.value`. This is okay most times, but if the update of `props.value` takes a while(i.e. `props.value` is coming from the backend and we're pushing it to the backend and reloading), this doesn't update the UI until `onUpdate` finishes.

Added functionality to disable the edit/delete popover as well when using it as non-editable

## Implementation Description

 - Changed the logic that forces `localMessages` to `props.value` if there is any discrepancy to become an effect, only to trigger when `props.value` changes.
 - Added nonEditable property to make the item non-editable, meaning the edit/delete functionality will not be there.
 - Moved the color setting of the chat out of isEditable into its own property `didUserWriteMessage`
 
N/A
## How Has This Been Tested?

Storybook
## PR Checklist:

Before changes:

https://github.com/user-attachments/assets/53855fc2-d5b6-49ba-ab7e-52982ab2d8fa

After changes:

https://github.com/user-attachments/assets/9acaf4a8-1b9d-4655-8193-4b4f45f8f3b8

NOTE: There still seems to be an issue where if there are multiple changes(multiple messages sent) and the process to update is quite slow, there will be a lag behind actually getting the items committed(since the onUpdate fires multiple times for each change). Unfortunately, I don't know how to get around this, but I do know we have a `loadingComponent` added, which we can utilise on time-intensive requests.


packages/ui-components/src/chat/chat.tsx
```diff
@@ -104,6 +104,8 @@ export interface ChatProps extends InteractiveComponentProps<Message[]> {
     isPopup?: boolean;
     /** A component showing the loading state of the chat, it appears above the input area, when not loading the caller can set it to null */
     loadingComponent?: React.ReactNode;
+    /** Whether the user can edit/delete previous messages */
+    nonEditable?: boolean;
```
krzysztof-causalens, line 108: Would it be clearer to name `isHistoryEditable`? Prefixing with `is` follows the boolean naming convention and adding history emphasises that we can e.g. write new messages - it's just the history (i.e. past messages) which is not editable.

krzysztof-causalens (approved):

sam-causalens (approved):
LGTM

PR #351, opened 2024-07-25
Improvement: DO-3471 Added message actions to show for each message

jsumiguin-causalens:
## Motivation and Context

Added functionality to append actions to a message. Actions will be additional elements that will be rendered with the message if the user plans to add additional functionality to the message.

## Implementation Description

 - Added actions property to `Message`
 - Passed this to `Message` component to be rendered in the UI
 
## How Has This Been Tested?

Locally
## PR Checklist:

### DIFF
--- a/packages/ui-components/src/chat/chat.tsx
+++ b/packages/ui-components/src/chat/chat.tsx
@@ -104,8 +104,6 @@ export interface ChatProps extends InteractiveComponentProps<Message[]> {
     isPopup?: boolean;
     /** A component showing the loading state of the chat, it appears above the input area, when not loading the caller can set it to null */
     loadingComponent?: React.ReactNode;
-
-    /** Pass through of className property */
 }
 
 /**
@@ -217,7 +215,6 @@ function Chat(props: ChatProps): JSX.Element {
                         onDelete={onDeleteMessage}
                         value={message}
                         isEditable={didUserWriteMessage(message, props.activeUser)}
-                        messageActions={message.actions}
                     />
                 ))}
                 {props.loadingComponent}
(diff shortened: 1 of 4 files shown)


packages/ui-components/src/chat/chat.tsx
```diff
@@ -215,6 +217,7 @@ function Chat(props: ChatProps): JSX.Element {
                         onDelete={onDeleteMessage}
                         value={message}
                         isEditable={didUserWriteMessage(message, props.activeUser)}
+                        messageActions={message.actions}
```
krzysztof-causalens, line 220: The entire message is already passed in as `value` do we need the extra prop?

packages/ui-components/src/types.tsx
```diff
@@ -79,6 +81,8 @@ export interface Message {
     updated_at: string;
     /** User data of the person who wrote the message */
     user: UserData;
+
+    actions?: React.ReactNode[];
```
krzysztof-causalens, line 85: Missing docstring

packages/ui-components/src/chat/chat.tsx
```diff
@@ -104,6 +104,8 @@ export interface ChatProps extends InteractiveComponentProps<Message[]> {
     isPopup?: boolean;
     /** A component showing the loading state of the chat, it appears above the input area, when not loading the caller can set it to null */
     loadingComponent?: React.ReactNode;
+
+    /** Pass through of className property */
```
krzysztof-causalens, line 108: Leftover?

krzysztof-causalens (approved):
Looks good. Could you add an example message with action(s) to the storybook page for future reference of intended usage?

sam-causalens (approved):
Looks good once there's a storybook example 👍

PR #353, opened 2024-08-01
Improvement: DO-3494: Automatically show index columns for Table, Support duplicate column names, Fix invisible boolean values in Table cells

Roman-Kornev:
## Motivation and Context

- Previously, any `Table(data=DataVariable(df))` would not show the index column when rendered. 
- Furthermore, when using Table with non-unique column names, the table would crash with `Error: ServiceError: DataFrame columns must be unique for orient='records'.`
- Additionally, the table would not show any cells with a boolean value.

All of the above issues have been fixed. Any index columns are now automatically included in the table and pinned to the left. Both multi-index rows and hierarchical columns are also supported. The `dtype` information is propagated to the column level, automatically rendering all supported types, including `DateTime`.

Other small issues fixed:
- Table `columns` field is now properly marked as optional.

## Implementation Description

In order to not break any existing APIs relying on `orient='records'` format, which omits the index information, the DataFrame is pre-processed internally by including a prefix `__index__{i}__` or `__col__{i}__` to a all columns and sent over the network.

When the table is rendered, the prefixes are stripped and used as `col_id` to map to the underlying data: `__col__2__name -> name`. This means that `getData()` call for column resolution now needs to happen when the columns are user-provided as well, but the table still shows the provided columns immediately. One small limitation is data returned by `onClickRow()` handler. Since it returns data as an object, any duplicate keys will be lost. This is preserved from before to not break any existing APIs.

Multi-index rows are included as new columns on the left, whereas hierarchical columns `('P', 'alpha'), ('Q', 'beta')]` are concatenated with `_` before rendering.

A new endpoint `/api/core/data-variable/${uid}/schema` is responsible for column type information. It it set any time data is retrieved. An optional `{ schema: boolean }` options field is added to `useDataVariable` to control whether the schema should be included in the response.

## How Has This Been Tested?

Tested locally. Fixed the tests and added test for internal format conversion.

## PR Checklist:

image

### Duplicate column names:
```py
df = pd.DataFrame({
        'A': [1, 2, 3],
        'B': [1.1, 2.2, 3.3]
    })
df2 = pd.concat([df, pd.DataFrame({
        'A': ['x', 'y', 'z'],
        'B': [1.1, 2.2, 3.3]
    })], axis=1)
df2
```
BEFORE:
image
AFTER:
image

### Multi-index rows and columns
```py
index = pd.MultiIndex.from_product([['A', 'B'], ['X', 'Y', 'Z']], names=['letter', 'symbol'])
columns = pd.MultiIndex.from_product([['P', 'Q'], ['alpha', 'beta']], names=['group', 'subgroup'])

df = pd.DataFrame(np.random.randn(6, 4), index=index, columns=columns)
df
```
BEFORE:
image
AFTER:
image


packages/dara-components/js/common/table/table.tsx
```diff
@@ -346,23 +364,67 @@ function Table(props: TableProps): JSX.Element {
...
+                );
+
+                const columnsWithoutGeneratedIndex = columns.filter((col) => col !== INDEX_COL);
+                let processsedColumns: ColumnProps[];
+                if (columnsProp) {
+                    // Prop provided, parse the columns provided and map them to the columns from data
+                    // Limitation: If there are columns with duplicate names, data from only one of them will be shown
+                    const reverseColumnIdMap = Object.fromEntries(
+                        columnsWithoutGeneratedIndex.map((col) => [
+                            extractColumnLabel(col, col.startsWith('__index__')),
```
krzysztof-causalens, line 397: I think we have a constant defined for the `__index__` prefix, we should probably use it everywhere instead of the magic string?

packages/dara-components/js/common/table/table.tsx
```diff
@@ -346,23 +364,67 @@ function Table(props: TableProps): JSX.Element {
...
-                setResolvedColumns(getColumnProps(columnsWithoutIndex));
+                const fieldTypes = Object.fromEntries(
+                    dataset.schema.fields.flatMap((field) => {
+                        const key = Array.isArray(field.name) ? field.name.join('_') : field.name;
+                        return [[key, { type: field.type }]];
+                    })
+                );
+
+                const columnsWithoutGeneratedIndex = columns.filter((col) => col !== INDEX_COL);
+                let processsedColumns: ColumnProps[];
```
krzysztof-causalens, line 391: Typo
```suggestion
                let processedColumns: ColumnProps[];
```

krzysztof-causalens (approved):
Looks good, nice job! This change needs a changelog entry and then it should be good to go once CI passes. 

I think there's room to improve the auto inferred columns in the future, now that we know for sure the dtypes we can e.g. add corresponding filter types to the columns. Happy for this to be done separately of course.

Regardless I think this change should be released in a new minor so it doesn't affect current internal releases and we can test it more extensively and see if people are happy with the behavior.

PR #354, opened 2024-08-05
Fix: DO-3494: Make Table schema cache_key a query param

Roman-Kornev:
## Motivation and Context

The param was being sent as a body which is not allowed for a GET request.

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/shared/interactivity/data-variable.tsx
+++ b/packages/dara-core/js/shared/interactivity/data-variable.tsx
@@ -168,7 +168,7 @@ async function fetchDataVariableSchema(
     extras: RequestExtras,
     cacheKey?: string
 ): Promise<DataResponse['schema']> {
-    const queryString = new URLSearchParams({ cache_key: cacheKey }).toString();
+    const queryString = cacheKey ? new URLSearchParams({ cache_key: cacheKey }).toString() : '';
     const response = await request(
         `/api/core/data-variable/${uid}/schema?${queryString}`,
         { method: HTTP_METHOD.GET },
--- a/packages/dara-core/js/shared/interactivity/data-variable.tsx
+++ b/packages/dara-core/js/shared/interactivity/data-variable.tsx
@@ -168,9 +168,10 @@ async function fetchDataVariableSchema(
     extras: RequestExtras,
     cacheKey?: string
 ): Promise<DataResponse['schema']> {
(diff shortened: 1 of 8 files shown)


packages/dara-core/tests/python/test_derived_data_variable.py
```diff
@@ -343,7 +343,9 @@ def func(a: int, data=TEST_DATA):
             json={'filters': {'column': 'col1', 'value': 2}, 'cache_key': cache_key, 'ws_channel': 'test_channel'},
             headers=AUTH_HEADERS,
         )
-        assert data_second_response.json() == df_convert_to_internal(unfiltered[unfiltered['col1'] == 2]).to_dict(orient='records')
+        assert data_second_response.json() == df_convert_to_internal(unfiltered[unfiltered['col1'] == 2]).to_dict(
```
krzysztof-causalens, line 346: Do we not have a test making a call for the schema with a cache key? It would be nice if a test caught an issue like that.

packages/dara-core/js/shared/interactivity/data-variable.tsx
```diff
@@ -168,9 +168,10 @@ async function fetchDataVariableSchema(
     extras: RequestExtras,
     cacheKey?: string
 ): Promise<DataResponse['schema']> {
+    const queryString = new URLSearchParams({ cache_key: cacheKey }).toString();
     const response = await request(
-        `/api/core/data-variable/${uid}/schema`,
-        { body: JSON.stringify({ cache_key: cacheKey }), method: HTTP_METHOD.GET },
+        `/api/core/data-variable/${uid}/schema?${queryString}`,
```
krzysztof-causalens, line 173: On the backend do we need to ensure the uriencoding is reverted? e.g. equivalent of decodeURIComponent. In case there's a mismatch this can cause it to miss the cache entry on the backend

packages/dara-core/tests/python/test_derived_data_variable.py
Roman-Kornev: Right, totally forgot about DDVs and that they use a different store. Added implementation and checked those in the test.

packages/dara-core/js/shared/interactivity/data-variable.tsx
Roman-Kornev: FastAPI automatically decodes all url-encoded query params

PR #355, opened 2024-08-06
Fix: DO-3494: Strip Table column prefix when filtering and sorting

Roman-Kornev:
## Motivation and Context

Due to the fact that Table column conversion happens as the last step, all filtering and sorting queries would try to operate on the raw data using the internal column names.

## Implementation Description

Rather than changing the API and forcing all transformations to happen on the converted data, the prefix is stripped altogether. The only downside is potentially in the future when we add automatic column filters: when trying to filter/sort non-unique column names - in that case the filtering would apply to all matching column names instead of the requested column. Right now it is not possible to specify column filters for duplicate column names anyway.

## How Has This Been Tested?

Added a pagination ordering test, added more testcases for filtering.

## PR Checklist:


krzysztof-causalens (approved):
Nice find, should be good enough for now

PR #356, opened 2024-08-07
Fix: DO-3494: Fix race conditon in registry entry registration

Roman-Kornev:
## Motivation and Context

When an entry is not found in the registry, we execute the handler to get the result and register it. While we execute the handler, something else might already register that specific entry, so before explicitly registering it again and getting an error we check it and only register it if it's still missing. Since registry is a dict, both `.get` and `.register` are atomic.

## PR Checklist:


krzysztof-causalens (approved):
LGTM, nice catch

PR #358, opened 2024-08-14
Fix: DO-3605 code viewer code area overflows container

patricia-causalens:
## Motivation and Context

Noticed that code viewer component could overflow its container if width was set to it

## Implementation Description

Fixed so that it can scroll horizontally when needed.

Also when updating the changelogs noticed that all of the ui packages did not have their NEXT replaced. When looking into it this was due to the changelog being placed one layer deeper than other dara packages. I moved them so that these are all the same now.

None

## How Has This Been Tested?

Locally on storybook and have also tested by running tooling command that ui packages changelogs are updated

## PR Checklist:

#### BEFORE
Screenshot 2024-08-14 at 15 19 19

#### AFTER
Screenshot 2024-08-14 at 15 18 40


tamara-causalens (approved):
Nice find!

krzysztof-causalens (approved):

PR #359, opened 2024-08-15
Fix: DO-3618 KeyError in LRU cache

krzysztof-causalens:
## Motivation and Context

Issue reported in internal product where in some rare circumstances the LRU code throws a KeyError when evicting an element from the cache store. 
The exact scenario is unknown, one hypothesis is that there is a threading problem where the LRU uses an async lock which is bypassed by accessing the store in a different thread.

The fix for now is just using `pop` instead of `del` to not cause a crash if the deletion fails.

## How Has This Been Tested?

Existing cache store tests passing

## PR Checklist:


patricia-causalens (approved):
LGTM

sam-causalens (approved):

PR #363, opened 2024-08-23
Fix: DO-3658: Fix Table non-unique index values, fix sorting by unnamed index

Roman-Kornev:
## Motivation and Context

-   Fixed a crash in `Table` pagination where rows containing non-unique index values would cause a slicing error, as rows were paginated by DataFrame index instead of `iloc` row ordering.
-   Fixed an issue in `Table` where sorting by an unnamed index would not work, as the default `__index__` unnamed index column does not exist in the original DataFrame.

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,11 @@
 title: Changelog
 ---
 
+## NEXT
+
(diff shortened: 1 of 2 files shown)


patricia-causalens (approved):
Good catch, LGTM!

krzysztof-causalens (approved):
LGTM

PR #364, opened 2024-08-23
Fix: DO-3637: Fix rendering of tables from DerivedDataVariables

Roman-Kornev:
## Motivation and Context

Fixed a crash in `Table` when rendering the result of a `DerivedDataVariable` due to a missing `cache_key`.

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/shared/interactivity/data-variable.tsx
+++ b/packages/dara-core/js/shared/interactivity/data-variable.tsx
@@ -286,8 +286,8 @@ export function useFetchDerivedDataVariable(
             // For derived data variables count can only be fetched when task is not running so we have to make the request here
             // As the total count could have changed because of the underlying DV changing
             const [totalCount, schema = null] = await Promise.all([
-                fetchDataVariableCount(variable.uid, extras, mergedFilters),
-                ...(withSchema ? [fetchDataVariableSchema(variable.uid, extras)] : []),
+                fetchDataVariableCount(variable.uid, extras, mergedFilters, dvValue.cache_key),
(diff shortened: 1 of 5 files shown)


packages/dara-core/js/shared/interactivity/data-variable.tsx
```diff
@@ -286,8 +286,8 @@ export function useFetchDerivedDataVariable(
             // For derived data variables count can only be fetched when task is not running so we have to make the request here
             // As the total count could have changed because of the underlying DV changing
             const [totalCount, schema = null] = await Promise.all([
-                fetchDataVariableCount(variable.uid, extras, mergedFilters),
-                ...(withSchema ? [fetchDataVariableSchema(variable.uid, extras)] : []),
+                fetchDataVariableCount(variable.uid, extras, mergedFilters, dvValue.cache_key),
+                ...(withSchema ? [fetchDataVariableSchema(variable.uid, extras, dvValue.cache_key)] : []),
             ] as const);
```
krzysztof-causalens, line 291: Do we not have any tests that ensure that at least we make the right request to the backend with the data? Might help prevent mistakes here in the future

krzysztof-causalens (approved):
Nice catch!

sam-causalens (approved):

packages/dara-core/js/shared/interactivity/data-variable.tsx
Roman-Kornev: Good point. We had a test, but it was not checking for the `cache_key` presense. Overrode the handler for DDV that specifically checks for the key there.

Also, some types were missing the schema options flag, fixed those as well.

PR #366, opened 2024-09-04
Fix: DO-3770 add min width to causal graph viewer

patricia-causalens:
## Motivation and Context

In some cases where you had a `CausalGraphViewer` next to a `Card` it would render with 0 width, making it confusing for users to understand that it was there and how to fix it.

## Implementation Description

Similarly to how we do for height, added a `min-width` of `100px`, just so it renders with something, from there it is easy for a user to update this value if needed.

None

## How Has This Been Tested?

Locally

## PR Checklist:

### DIFF
--- a/packages/ui-causal-graph-editor/changelog.md
+++ b/packages/ui-causal-graph-editor/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Added a `min-with` to `CausalGraphEditor`.
+
 ## 1.12.7
 
 -   Fix an issue where body's `line-height` was not consistent across packages.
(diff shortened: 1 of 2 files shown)


packages/ui-causal-graph-editor/changelog.md
```diff
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-   Added a `min-with` to `CausalGraphEditor`.
```
krzysztof-causalens, line 7:
```suggestion
-   Added a `min-width` to `CausalGraphEditor`.
```

krzysztof-causalens (approved):

PR #367, opened 2024-09-09
Improvement: DO-3720 graph layout calculation performance

krzysztof-causalens:
## Motivation and Context

This PR introduces the long-awaited change to offload graph layout computations to a webworker instead of running them in the main thread. 

The setup to get this working when distributed requires switching our build pipeline for the package to use `vite build` instead of just `tsc`, to have Vite transform the worker into its inline form (for simplicity) in the shipped code. Otherwise supporting all the different modes (UMD, prod, dev Dara builds) would be very difficult.

Other changes:
- fixed memory leak in node grouping caused by extra node textures being created unnecessarily
- fixed node grouping lagging - had to remove the border to reduce number of textures created but it drastically improved perf so should be good enough

## Implementation Description

The webworker setup is quite complicated due to our usage of different bundlers - Storybook uses Webpack while Dara uses Vite (quite old version).
I attempted a quick Storybook migration to Vite but was unable to get it to work - stories rendered but the worker was not launching. Unsure what it was caused by, didn't find any solutions in a little bit so abandoned the idea for now.

The current workaround is to have the source code reflect the needs to make it work in Dara (UMD, prod, dev). I left a commented out piece of code to make it work in storybook. There is a comment about it in the README. 

The setup I've chosen uses the `inline` worker mode in Vite (literally inlines base64 worker code into the bundle) as it's the easiest to bundle.

- comlink - library to make working with web workers much easier

## How Has This Been Tested?

Tested manually in a test app in the monorepo: 
- UMD mode
- dev mode
- prod mode

Tested in storybook

Tested in a separate Dara project (by copying changes to node_modules) in dev and prod mode. 

## PR Checklist:

Tested in Dara, no issues with number of workers - 16 should be the limit of WebGL contexts on a single page. I tried rendering 100 graphs and workers are created fine, though we run into the error "WARNING: Too many active WebGL contexts. Oldest context will be lost.".

Ran dara in Dev, UMD and Prod mode.

https://github.com/user-attachments/assets/3d06c3fe-1d69-4869-9522-35ce223f77f6


packages/ui-causal-graph-editor/src/shared/rendering/grouping/group-container-object.tsx
```diff
@@ -109,22 +109,19 @@ export class GroupContainerObject extends PIXI.EventEmitter<(typeof MOUSE_EVENTS
         (groupContainerGfx.hitArea as PIXI.Rectangle).height = height;
 
         // Get/create rectangle texture
-        const rectangleTexture = textureCache.get(createKey(GROUP_RECTANGLE, minX, maxX, minY, maxY), () => {
+        const rectangleTexture = textureCache.get(createKey(GROUP_RECTANGLE), () => {
```
krzysztof-causalens, line 112: This change fixes a memory leak. Previously this would create a new texture for each unique rectangle position.

Now there is no border at all so the rectangle texture is static and is then resized by the surrounding sprite which is much more performant.

patricia-causalens: Thank you!

Roman-Kornev:
Maybe it would be better to split the PR into two: the `tsc` change and the layout implementation? No need to do it now though.

krzysztof-causalens:
> Maybe it would be better to split the PR into two: the `tsc` change and the layout implementation? No need to do it now though.

Yeah you're right could've done that, it's just that I was tweaking the tsconfig to get it to output declarations, and as I was trying the vite builder it didn't work unless I changed to use verbatimModuleSyntax so I bundled it. In the end I didn't switch to the vite builder so the changes could be moved but they were already done by that point so I kept them in.

krzysztof-causalens:
> LGTM.
> 
> I'm curious, have you measured roughly what the overhead of moving the implementation to the worker is? Passing all those messages around has a cost. I would imagine it's not much as the bulk of the processing is CPU heavy.

I'll measure it and post the details here, good point. From empirical evidence it's not noticeable and feels very fast.

packages/ui-causal-graph-editor/src/shared/graph-layout/worker/worker.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Callbacks returned by the layout computation to be invoked by requests
+ */
+const LAYOUT_CALLBACKS: LayoutComputationCallbacks = {} as const;
+
+/**
+ * Compute the layout for the given graph using the provided layout parameters.
+ *
+ * @param layoutParams layout parameters
+ * @param graph the graph to apply the layout to
```
patricia-causalens, line 47: serializedGraph instead of graph

packages/ui-causal-graph-editor/src/types.tsx
```diff
@@ -205,6 +205,14 @@ export interface SimulationAttributes {
 }
 export type SimulationGraph = AbstractGraph<SimulationNode, SimulationEdge, SimulationAttributes>;
 
+// Defines the Layering algorithms supported by PlanarLayout
+export enum LayeringAlgorithm {
```
patricia-causalens, line 209: Since this is defined in types now do you think we should call it PlanarLayeringAlgortithm or something more specific to make it clear it only works for that?

patricia-causalens (changes requested):
Hi Krzys,

Great work on this! Just some very minor things from the code side.

I was playing with it locally and noticed that my storybook can get stuck when adding edges in the very large graphs. When adding nodes it does go in the calculating layout state, so wondering if maybe we might be missing some edge case (edge case with edges?!) here? From what I remember the graph is re-rendered when an edge is added, and in some cases it does recalculate the layout e.g. for planar. But don't remember if it is the same for fcose which is where I saw it happening.

packages/ui-causal-graph-editor/src/shared/editor-overlay/zoom-prompt.tsx
```diff
@@ -20,6 +20,7 @@ const PromptWrapper = styled.div`
     align-items: center;
 
     min-width: 0;
+    height: 40px;
```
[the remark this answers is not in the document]
krzysztof-causalens: This width is also applied to the top-level buttons, we probably don't want to scale the button size with font size.

packages/ui-causal-graph-editor/src/graph-viewer/causal-graph-editor.tsx
```diff
@@ -760,6 +791,9 @@ function CausalGraphEditor({ requireFocusToZoom = true, ...props }: CausalGraphE
                                 </>
                             }
                             topLeft={<RecalculateLayoutButton onResetLayout={resetLayout} />}
+                            loadingIndicator={
+                                isLayoutComputing && <LayoutSpinner text="Calculating layout..." size="24px" />
```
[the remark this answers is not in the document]
krzysztof-causalens: Same reason here, the size I picked to align with the size with the top buttons so it doesn't look out of place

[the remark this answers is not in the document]
patricia-causalens: Would it be better to have some height like the banner so the font-size changes but it stays aligned? Because the text in the banner will change size, or maybe that shouldn't change either?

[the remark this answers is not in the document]
krzysztof-causalens: The font-size is 1.2rem atm, I'll probably just hard-code it for now to match the size.

krzysztof-causalens:
Fixed bugs found:
- when symbols were hidden by zooming out, they would not reappear
- cycle detection was running in non-DAG mode which would break if the graph had cycles already causing it to run into an infinite loop (and hanging until it reached the max array length then crashing)

krzysztof-causalens:
> LGTM.
> 
> I'm curious, have you measured roughly what the overhead of moving the implementation to the worker is? Passing all those messages around has a cost. I would imagine it's not much as the bulk of the processing is CPU heavy.

Re: overhead. It's difficult to measure as timers don't work well cross-thread. From manual timestamping the overhead seems to be fairly low, the most I've seen was 500ms on a very large graph but at this point the calculation itself took 20+ seconds so in comparison it's minimal. There is no real alternative either way as we have to offload it.
The overhead for smaller graphs is unnoticeable so there is no reason to do any more complex logic (such as running it in-thread in some cases).

packages/ui-causal-graph-editor/src/shared/graph-layout/worker/spring.ts
```diff
@@ -317,7 +327,9 @@ export default function compute(
             simulation.alpha(0.1).alphaTarget(0);
         },
         onMove: (nodeId: string, x: number, y: number) => {
+            console.log('move', nodeId);
```
patricia-causalens, line 330: left over console.logs

packages/ui-causal-graph-editor/src/shared/rendering/edge/symbols.tsx
```diff
@@ -34,6 +34,7 @@ export function createSideSymbol(
     tint: number,
     bgTint: number
 ): PIXI.Graphics {
+    console.log('creating symbol', style);
```
patricia-causalens, line 37: Another one

patricia-causalens (approved):
Thanks so much for looking at all these bugs! Just some leftover console.logs then it is good to go! :)

PR #368, opened 2024-09-13
Fix: DO-3745 table display does not format datetime columns correctly

patricia-causalens:
## Motivation and Context

User was passing `dataframe.describe()` as the data to a `Table` component, in this case one of their columns was of type `datime64`. Instead of showing the datetime values formatted as one would expect when printing this information, the table instead showed the dates as timestamps.

## Implementation Description

Looking closer at the issue, the `describe` method in pandas for `datetime64` columns returns a column of type `object`. This is because `object` is the more generic type, the resulting column is a mixture of datetime and number values. Due to this the Table component was not able to infer the correct types and when the data was serialized Timestamp(...) became just the timestamp string.

What we can do to address the issue is the same that pandas does in their `__repr__` for this type of column, which is to convert each value to a str. 

This transformation was implemented in the endpoint the table uses to fetch the data, so that it shows a correct representation, but the `DataVariable`'s value remains unaffected.

None

## How Has This Been Tested?

Changes tested on internal product where this issue was found

## PR Checklist:

BEFORE:
Screenshot 2024-09-13 at 17 29 21

AFTER:
Screenshot 2024-09-13 at 17 28 50

### DIFF
--- a/packages/dara-core/dara/core/interactivity/data_variable.py
+++ b/packages/dara-core/dara/core/interactivity/data_variable.py
@@ -212,6 +212,11 @@ class DataVariable(AnyDataVariable):
 
         if entry.data is not None:
             filtered_data, count = apply_filters(entry.data, coerce_to_filter_query(filters), pagination)
+            if filtered_data is not None:
+                for col in filtered_data.columns:
+                    if filtered_data[col].dtype == 'object':
+                        # We need to convert all values to string to avoid issues with displaying data in the Table component, for example when displaying datetime and number objects in the same column
+                        filtered_data[col] = filtered_data[col].apply(str)
             data = filtered_data
             # Store count for given filters and schema
             await asyncio.gather(
--- a/packages/dara-core/dara/core/interactivity/data_variable.py
+++ b/packages/dara-core/dara/core/interactivity/data_variable.py
@@ -182,6 +182,7 @@ class DataVariable(AnyDataVariable):
         store: CacheStore,
         filters: Optional[Union[FilterQuery, dict]] = None,
         pagination: Optional[Pagination] = None,
+        format_for_display: bool = False,
     ) -> Optional[DataFrame]:
         """
         Get the value of this DataVariable.
@@ -212,7 +213,7 @@ class DataVariable(AnyDataVariable):
 
         if entry.data is not None:
             filtered_data, count = apply_filters(entry.data, coerce_to_filter_query(filters), pagination)
-            if filtered_data is not None:
+            if format_for_display and filtered_data is not None:
                 for col in filtered_data.columns:
                     if filtered_data[col].dtype == 'object':
                         # We need to convert all values to string to avoid issues with displaying data in the Table component, for example when displaying datetime and number objects in the same column
--- a/packages/dara-core/dara/core/interactivity/data_variable.py
+++ b/packages/dara-core/dara/core/interactivity/data_variable.py
@@ -214,10 +214,11 @@ class DataVariable(AnyDataVariable):
         if entry.data is not None:
             filtered_data, count = apply_filters(entry.data, coerce_to_filter_query(filters), pagination)
             if format_for_display and filtered_data is not None:
+                filtered_data = filtered_data.copy()
                 for col in filtered_data.columns:
                     if filtered_data[col].dtype == 'object':
                         # We need to convert all values to string to avoid issues with displaying data in the Table component, for example when displaying datetime and number objects in the same column
-                        filtered_data[col] = filtered_data[col].apply(str)
+                        filtered_data.loc[:, col] = filtered_data[col].apply(str)
             data = filtered_data
             # Store count for given filters and schema
             await asyncio.gather(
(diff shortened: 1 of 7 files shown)


packages/dara-core/dara/core/interactivity/data_variable.py
```diff
@@ -212,6 +212,11 @@ async def get_value(
 
         if entry.data is not None:
             filtered_data, count = apply_filters(entry.data, coerce_to_filter_query(filters), pagination)
+            if filtered_data is not None:
```
krzysztof-causalens, line 215: The `get_value` method is also invoked when resolving data variable values as e.g. a DerivedVariable dependency.

Perhaps this can just be a flag on this method that we would set when calling it within the endpoint? We can then add further post processing for display that won't affect the raw data for other purposes.

packages/dara-core/dara/core/interactivity/data_variable.py
```diff
@@ -212,6 +212,11 @@ async def get_value(
...
+                for col in filtered_data.columns:
+                    if filtered_data[col].dtype == 'object':
+                        # We need to convert all values to string to avoid issues with displaying data in the Table component, for example when displaying datetime and number objects in the same column
+                        filtered_data[col] = filtered_data[col].apply(str)
```
krzysztof-causalens, line 219: This also needs to happen for DerivedDataVariables.

krzysztof-causalens (changes requested):
Looks good but the current implementation will also affect DerivedVariables where we should preserve the original data

packages/dara-core/dara/core/interactivity/derived_data_variable.py
```diff
@@ -297,6 +298,11 @@ async def get_data(
 
         # Run the filtering
         data = await cls._filter_data(data, count_cache_key, data_entry, store, filters, pagination)
+        if format_for_display and data is not None:
+            for col in data.columns:
+                if data[col].dtype == 'object':
+                    # We need to convert all values to string to avoid issues with displaying data in the Table component, for example when displaying datetime and number objects in the same column
+                    data[col] = data[col].apply(str)
```
krzysztof-causalens, line 305: One thing I just thought about - is this operating on a copy of the data or the original one? Could you double check whether that's the case, e.g. if you have a Table displaying a DataVariable/DerivedDataVariable and also a different e.g. DerivedVariable consuming the data, is it not affected?

packages/dara-core/dara/core/internal/routing.py
```diff
@@ -332,6 +332,7 @@ async def get_data_variable(
                     store,
                     body.filters,
                     Pagination(offset=offset, limit=limit, orderBy=order_by, index=index),
+                    True,
```
krzysztof-causalens, line 335: Could you pass the boolean as a kwarg in both places, would be much more readable than having to look up what this `True` means here

packages/dara-core/dara/core/interactivity/data_variable.py
```diff
@@ -212,6 +213,11 @@ async def get_value(
 
         if entry.data is not None:
             filtered_data, count = apply_filters(entry.data, coerce_to_filter_query(filters), pagination)
+            if format_for_display and filtered_data is not None:
+                for col in filtered_data.columns:
+                    if filtered_data[col].dtype == 'object':
+                        # We need to convert all values to string to avoid issues with displaying data in the Table component, for example when displaying datetime and number objects in the same column
+                        filtered_data[col] = filtered_data[col].apply(str)
```
Roman-Kornev, line 220: This should also help with cases where a cell contains a non-primitive value, like a list

Roman-Kornev (approved):
LGTM, thanks for this

krzysztof-causalens (approved):

packages/dara-core/dara/core/interactivity/data_variable.py
```diff
@@ -214,10 +214,11 @@ async def get_value(
         if entry.data is not None:
             filtered_data, count = apply_filters(entry.data, coerce_to_filter_query(filters), pagination)
             if format_for_display and filtered_data is not None:
+                filtered_data = filtered_data.copy()
                 for col in filtered_data.columns:
                     if filtered_data[col].dtype == 'object':
                         # We need to convert all values to string to avoid issues with displaying data in the Table component, for example when displaying datetime and number objects in the same column
-                        filtered_data[col] = filtered_data[col].apply(str)
+                        filtered_data.loc[:, col] = filtered_data[col].apply(str)
```
Roman-Kornev, line 221: Will this still work for `MultiIndex` rows/`MultiIndex` columns?

patricia-causalens:
When testing in internal product found out that DataSlicer would error when rendering as it couldn't find the filter button in the registry. Have updated to be exported here as well

packages/dara-core/dara/core/interactivity/data_variable.py
patricia-causalens: I have tested with MultiIndex and as far as I can tell seems to work without any issues

PR #369, opened 2024-09-25
Fix: DO-3828 table showing wrong dates

patricia-causalens:
## Motivation and Context

User found that with their dataset the timestamp column showed with dates in 1970s whereas when printing with pandas these were correctly inferred to be in 2023.

## Implementation Description

It seems that in this case the column had timestamps in seconds. When we tried to get the corresponding dates with `formatISO` we assumed the timestamp is given in ms. I have updated this to now consider that the timestamp might be in seconds or nanoseconds and convert accordingly.

Limitations with this approach:
- If dates go beyond the year 33658 then dates would be wrongly inferred as ms will reach 1e15 and seconds reach the ms range. This seems unlikely for vast majority of what might be fed to the table component, so seemed like an ok assumption.

- We also don't take into account microseconds. I don't think it is that commonly used, but can add if we think it might be.

## How Has This Been Tested?

Locally with some test data

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,9 @@
 title: Changelog
 ---
 
+## NEXT
+-   Fixes an issue where `Table` would sometimes convert timestamps in seconds as if they were in milliseconds.
+
 ## 1.12.3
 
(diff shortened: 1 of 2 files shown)


krzysztof-causalens (approved):
LGTM, nice find

PR #370, opened 2024-09-30
Fix: DO-3869 input component bug

patricia-causalens:
## Motivation and Context

User found that if they type `10.` and then tried to backspace, then it didn't go to `10`

## Implementation Description

The issue was found that when we updated the input to `10.`, we keep the value at `10`, so when the backspace is pressed the value does go to `10`, but since the value hadn't changed then the useEffect was not triggered and so the input was not updated.

None

## How Has This Been Tested?

Locally on storybook and by adding a test

## PR Checklist:

### DIFF
--- a/packages/ui-components/src/numeric-input/numeric-input.spec.tsx
+++ b/packages/ui-components/src/numeric-input/numeric-input.spec.tsx
@@ -32,6 +32,16 @@ function RenderNumericInput(props: NumericInputProps): JSX.Element {
     );
 }
 
+const MockControlledNumericInput = (): JSX.Element => {
+    const [value, setValue] = React.useState();
+
+    const handleChange = (newValue): void => {
+        setValue(newValue);
+    };
+
+    return <RenderNumericInput value={value} onChange={handleChange} />;
+};
+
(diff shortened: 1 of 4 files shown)


krzysztof-causalens (approved):
LGTM, nice catch :)

packages/ui-components/src/numeric-input/numeric-input.spec.tsx
```diff
@@ -95,6 +105,17 @@ describe('Numeric Input', () => {
...
     });
 
+    it('should erase . if decimal value was not fully entered', async () => {
+        const { getByRole } = render(<MockControlledNumericInput />);
+        const input = getByRole('textbox', { hidden: true });
+
+        await userEvent.type(input, '10.', { delay: 1 });
+        expect(input).toHaveValue('10.');
+
+        userEvent.keyboard('{backspace}');
```
krzysztof-causalens, line 115: Missing await?

patricia-causalens: I had it initially but ide corrected saying it is not needed here

PR #373, opened 2024-10-08
Fix: DO-3936 bar plot not picking up default dara styling with plotly

patricia-causalens:
## Motivation and Context

User found that in their bar chart that uses plotly express that the bar colors appeared purple, whereas in normal plotly bar it appeared with our primary blue color.

## Implementation Description

It seems that when using plotly express, the bar colors are passed to the marker property in the `data` of the figure. 

Since we don't know what `data` a figure will have, on the `Plotly` component we set our colors as part of the `layout` property of the figure. 

Based on perusing through plotly's code, when plotly express's bar method is called it looks for the `color` property set in the currently active template. Which means that it gets the default first color available in plotly's default theme (the purple we see).

To address this I have added at the top level so that when `Plotly` component is imported we also set the default template to use the layout changes we define in plotly itself.

Caveat here is that if you set the dark theme with `set_default_plotly_theme`, for the plotly espress bar chart it will show the light version of primary color instead of the dark one. I am not sure what we could do about this. resetting the template in this function did not work. Since it is a very specific and uncommon scenario we can come back to this if it becomes an issue in the future. Users can still overwrite and set the color they want to the bars so there is an easy workaround.

## How Has This Been Tested?

Locally in light and dark modes

## PR Checklist:

BEFORE:
Screenshot 2024-10-08 at 10 58 50

AFTER:
Screenshot 2024-10-08 at 10 58 25

Screenshot 2024-10-08 at 10 58 07

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -5,6 +5,7 @@ title: Changelog
 ## NEXT
 -   Fixed an issue where `Table` would sometimes convert timestamps in seconds as if they were in milliseconds.
 -   Fixed an issue where `CheckboxGroup` could not have an `undefined` value.
+-   Fixed an issue where `Plotly` in some cases would not pick up colors from dara theme by default.
 
 ## 1.12.3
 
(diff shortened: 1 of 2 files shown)


krzysztof-causalens (approved):
LGTM, nice find. Should be good enough for now, let's see how it works for people

PR #374, opened 2024-10-11
Fix: DO-3956 causal graph viewer not updating nodes when graph is updated

patricia-causalens:
## Motivation and Context

Multiple users reported that sometimes the nodes and edges of their graph would disappear. But after interacting with the graph they magically appeared again.

## Implementation Description

The issue was found to be related to [this issue](https://github.com/pixijs/pixijs/issues/10735) on Pixijs.

When a node was added or the graph was resized, the node could temporarily be positioned outside the visible frame. This occurred either because a newly created node did not yet have a defined position or because resizing the graph to a smaller size pushed existing nodes outside the frame. As a result, the nodes were being prematurely culled and rendered invisible before reaching their final render state.

The fix was added to `updateLayout`, as this method is called in both scenarios where the issue occurred. The solution forces a re-render, ensuring the latest frame is used for culling. 

This PR also adds edge symbols and labels to be culled. 

None

## How Has This Been Tested?

Locally on storybook and with the example in which this was initially spotted with.

## PR Checklist:

BEFORE:
before node fix

AFTER:
node fix

### DIFF
--- a/packages/ui-causal-graph-editor/src/shared/rendering/engine.tsx
+++ b/packages/ui-causal-graph-editor/src/shared/rendering/engine.tsx
@@ -1861,7 +1861,9 @@ export class Engine extends PIXI.EventEmitter<EngineEvents> {
             // which fires events so just in case fire one here
             this.onLayoutComputationDoneBound();
 
-            // Without this in some instances when the graph resizes or nodes are added they can be culled making them invisible
+            // Force re-render to ensure nodes are positioned correctly within the frame
+            // after a new node is added or the graph is resized. This prevents nodes
+            // from being culled prematurely by ensuring the latest frame is used for culling.
             setTimeout(() => {
                 this.viewport.dirty = true;
             }, 300);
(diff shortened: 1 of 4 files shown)


packages/ui-causal-graph-editor/src/shared/rendering/engine.tsx
```diff
@@ -1860,6 +1860,13 @@ export class Engine extends PIXI.EventEmitter<EngineEvents> {
             // ensure the callback is invoked - the layout might be custom/not go through the worker
             // which fires events so just in case fire one here
             this.onLayoutComputationDoneBound();
+
+            // Force re-render to ensure nodes are positioned correctly within the frame
+            // after a new node is added or the graph is resized. This prevents nodes
+            // from being culled prematurely by ensuring the latest frame is used for culling.
+            setTimeout(() => {
+                this.viewport.dirty = true;
+            }, 300);
```
krzysztof-causalens, line 1869: Have you tested if the timeout needs to be this large? It's quite visible so I'm wondering if we can get away with a shorter one.

Make sure it works for larger graphs too though in case it takes a bit longer to render.

Roman-Kornev (approved):

packages/ui-causal-graph-editor/src/shared/rendering/engine.tsx
patricia-causalens: Managed to reduce to 100 and seems to work fine, tried with differently sized graphs

krzysztof-causalens: Nice, should be much less noticable

krzysztof-causalens (approved):
LGTM

PR #375, opened 2024-10-11
Fix: DO-3941 custsup plotly with too many data points breaks with enumerate error

patricia-causalens:
## Motivation and Context

User was trying to render a Plotly figure on internal product that had over a million data points. This caused Plotly component to break with "too many properties to enumerate error"

## Implementation Description

Looking through the error the issue was found to be triggered by this line:
`Object.keys(variable).includes('uid')`

variable in this case included the plotly figure which due to its size had a substantial amount of properties.

The suggested fix is to use `hasOwnProperty` instead. This avoids the overhead of creating an array of keys and improves performance by checking only the necessary property.

None

## How Has This Been Tested?

By applying the change on internal product where the issue occurs

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/types/utils.tsx
+++ b/packages/dara-core/js/types/utils.tsx
@@ -18,6 +18,7 @@ import {
 export function isVariable<T>(variable: AnyVariable<T> | T): variable is AnyVariable<T> {
     return (
         variable &&
+        typeof variable == 'object' &&
         variable.hasOwnProperty('uid') &&
         variable.hasOwnProperty('__typename') &&
         (variable as { __typename: string }).__typename.includes('Variable')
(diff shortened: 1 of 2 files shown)


packages/dara-core/js/types/utils.tsx
```diff
@@ -18,8 +18,8 @@ import {
 export function isVariable<T>(variable: AnyVariable<T> | T): variable is AnyVariable<T> {
     return (
         variable &&
-        Object.keys(variable).includes('uid') &&
-        Object.keys(variable).includes('__typename') &&
+        variable.hasOwnProperty('uid') &&
+        variable.hasOwnProperty('__typename') &&
```
krzysztof-causalens, line 22: For extra speed you can probably also add a check of the type, i.e.

`variable && typeof variable == 'object' && variable.hasOwnProperty...`

krzysztof-causalens (approved):
LGTM, nice find!

packages/dara-core/js/types/utils.tsx
patricia-causalens: Good point, updated

PR #377, opened 2024-10-14
Fix: DO-3894 native plotly plots cannot change size

patricia-causalens:
## Motivation and Context

Issue found in internal product where when setting height on a figure layout that the resulting plot did not have correct height.

## Implementation Description

Internal product uses dara's `Plotly` component under the hood. The fix was aimed at improving this for both dara apps and the internal product. 

Since dara uses an Autosizer around plotly any height set within the component was being ignored and the autosizer value replaced it. In order to improve the behaviour to be able to set `height` as both part of the figure layout and prop to the component, I added a style to check if figure has this set and if so to use it instead. The order of preference becomes:
1. Height/width set to Plotly component
2. Height/width set to Plotly figure
3. Use autosizer

None

## How Has This Been Tested?

Locally on a dara app and by testing these changes in the internal product

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,7 +2,12 @@
 title: Changelog
 ---
 
+# NEXT
+
+-   Fixed an issue where `Plotly`'s figure native `height` and `width` were not obeyed by component.
+
 ## 1.13.1
+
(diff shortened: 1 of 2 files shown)


krzysztof-causalens (approved):

tamara-causalens (approved):
good catch!

PR #381, opened 2024-10-28
Fix: DO-3034 remove redundant data from refresh request

krzysztof-causalens:
## Motivation and Context

Issue noticed when testing refresh logic in internal product, seems to be working except there's unnecessary body data sent in the request from the original failing request.

## Implementation Description

By mistake I previously included the original request `RequestInit` data in the refresh request.

Once fixed I had to tweak the test setup to ensure `credentials: include` is set on the `fetch` calls in jest. In the browser the requests we make are same-origin so it's not necessary, but in the test runner they are treated as cross-origin so the extra setting is necessary. 

`@types/whatwg-fetch` because of the test setup above

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/auth/routes.py
+++ b/packages/dara-core/dara/core/auth/routes.py
@@ -19,15 +19,7 @@ from inspect import iscoroutinefunction
 from typing import Union, cast
 
 import jwt
-from fastapi import (
-    APIRouter,
-    BackgroundTasks,
-    Cookie,
-    Depends,
-    HTTPException,
-    Request,
-    Response,
-)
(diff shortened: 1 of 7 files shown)


packages/dara-core/dara/core/auth/routes.py
```diff
@@ -116,7 +108,6 @@ async def _revoke_session(response: Response, credentials: HTTPAuthorizationCred
 @auth_router.post('/refresh-token')
 async def handle_refresh_token(
     response: Response,
-    background_tasks: BackgroundTasks,
```
krzysztof-causalens, line 119: Was unused from a previous implementation

patricia-causalens (approved):
LGTM!

PR #386, opened 2024-12-03
Fix: DO-4205 loosen the croniter dependency

sam-causalens:
## Motivation and Context
Loosen the deps so dara can be installed with newer versions of airflow.

## Implementation Description
* We use the package minimally and the tests still run successfully on v2.x so we allow this version to be used as well as 1.x versions.
* Needed to allow the install of a newer version of airflow in another project.

no

## How Has This Been Tested?
covered by existing unit tests

## PR Checklist:


krzysztof-causalens (approved):
LGTM, thanks

PR #387, opened 2024-12-06
Fix: DO-4274 fix a couple of issues in the token refresh logic around websockets

sam-causalens:
## Motivation and Context
Issues seen in local dev with the messaging locking up, changes to the websocket client not getting propogated after a token refresh. This change also resolves the stale token issue on websockets where when there was a single tab the context var was set incorrectly.

## Implementation Description
* There was a misunderstanding in how the storage event worked that led to single open tabs not correctly refreshing their tokens when required. This is due to the event only being fired when storage is changed from another document. This meant that the session token in the useSessionToken hook was not correctly updated in the current tab.
* To address this the local subscribers field is now used again (it was left in a broken state, but not completely removed) and correctly pushes out updates to a key to all local subscribers immediately.
* Secondly, as we now have a new token update method for an existing websocket connection there is no longer a need to completely reconnect the websocket in such cases, we only need to update the token within. This is now the primary way for updating the token on the websocket client rather than recreating the whole connection each time we need to update the token.

No

## How Has This Been Tested?
Will try to add unit tests on Monday.

## PR Checklist:

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-  Fixed an issue with websocket token updates during an SSO driven token refresh. Tokens are now correctly updated in the websocket handler and in the UI.
+
 ## 1.14.2
 
 -   Internal (JS): remove redundant body from the token refresh request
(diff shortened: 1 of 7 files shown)


tamara-causalens (approved):

krzysztof-causalens (approved):
Oops, looks like that was my bad 🤦 Thanks for fixing! LGTM once there's some unit testing added

Roman-Kornev (approved):

PR #389, opened 2024-12-16
Fix: DO-4367 Loosen Jinja2 dep constraint

sam-causalens:
## Motivation and Context
* Other libs in our ecosystem (or ecosystem we wish to install into) have new versions that requireme jinja2 being > 3.1 so extend dara to allow it as it barely uses jinja and the api still works as before for what we do.

## Implementation Description
* Allow 3.1.x versions of Jinja2

No

## How Has This Been Tested?
Existing unit tests should catch any issues

## PR Checklist:


krzysztof-causalens (commented):
Have you tested this with an app? AFAIK the issue here is that `fastapi-vite` uses `Jinja` under the hood and they were using some API which was removed in some minor version. This isn't covered by unit tests, but if you've checked that everything works fine then LGTM.

sam-causalens:
> Have you tested this with an app? AFAIK the issue here is that `fastapi-vite` uses `Jinja` under the hood and they were using some API which was removed in some minor version. This isn't covered by unit tests, but if you've checked that everything works fine then LGTM.

Yeah, that seems to be the issue. I'll work on it tomorrow to try and resolve.

krzysztof-causalens:
> > Have you tested this with an app? AFAIK the issue here is that `fastapi-vite` uses `Jinja` under the hood and they were using some API which was removed in some minor version. This isn't covered by unit tests, but if you've checked that everything works fine then LGTM.
> 
> Yeah, that seems to be the issue. I'll work on it tomorrow to try and resolve.

FYI we've investigated this in the past, the solution is to fork it since the library is basically unmaintained (outside of their move to Pydantic v2). We had it planned, just never got around to it - should be fairly simple, it's a small library, just needs creating a repo, fixing the issue and publishing to PyPI.

krzysztof-causalens (approved):
LGTM if it works correctly with the fork, thanks

sam-causalens:
> LGTM if it works correctly with the fork, thanks

Yep looks to work well. I span up a local app to test and that worked too 👍

PR #391, opened 2024-12-20
Improvement: DO-4388 updated reference docs with defined variables

claudiawestby:
## Motivation and Context
often times in the reference docs we don’t assign our variables to a value we use them directly in the component ie:
```python
Input(value=Variable('test'))
```
99% of the time we would want to assign our variable like the following:
```python
value_var = Variable('test')
Input(value=value_var)
```
so we can actually get access to that value later on in our component/apps

updated the reference docs to avoid this bad practice so that the dara agent does not pick this habit up

## Implementation Description
updated some reference docs 

## PR Checklist:

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+-  Clarified `Variable` usage in the docs of some components.
+
 ## 1.14.3
 
 -   Added the `language` prop to the `CodeEditor` component to allow for setting the language of the code editor.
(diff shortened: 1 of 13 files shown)


Roman-Kornev (approved):
Is the aim to update all such usages? I suggest also updating `docs/getting-started/actions.md`, `components.md`, and `dara/core/interactivity/actions.py`, etc.

claudiawestby:
@Roman-Kornev for more advanced topics like actions, derived variables etc. i write my own heuristics in the agent (essentially no .md or .mdx files are actually used). it's easier this way to tweak and see what works - maybe after some time to prove that what we have in there is relatively stable we can port it over to their relevant sections in the documentation and just link the docs :) but would say that's a future task

PR #393, opened 2025-01-30
Fix: DO-4864 Resolve websocket not reconnecting correctly.

sam-causalens:
## Motivation and Context
Address Studio being broken after coming back to it after leaving the tab for a while.

## Implementation Description
* There were two issues causing this. The first was that we were raising HTTPExceptions rather than Websocket exceptions during the handshake and upgrade from http request to websocket. This caused the error to be swallowed and replaced with a generic error log and handshake failed response.
* Switching to using WebsocketException now gets properly handled and the logs now correctly show the 403 exception and logged detail rather than swallowing everything.
* The second part of the issue was that the websocket client on the JS side was getting stuck with a stale token if this happened whilst the computer was sleeping / tab sleeping. I suspect this was caused by updateToken trying to send a message on the disconnected socket and erroring out before updating the token on the class.
* To address this, the token on the class is now set first regardless of the state of the websocket connection. During initialization we also pull the latest token from the globalStore to ensure it's up to date if for some reason the token does get out of sync.

No

## How Has This Been Tested?
Tested in a local studio

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/internal/websocket.py
+++ b/packages/dara-core/dara/core/internal/websocket.py
@@ -14,6 +14,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+
 import asyncio
 import inspect
 import math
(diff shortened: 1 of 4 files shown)


packages/dara-core/dara/core/internal/websocket.py
```diff
@@ -405,7 +414,7 @@ async def ws_handler(websocket: WebSocket, token: Optional[str] = Query(default=
     :param token: The authentication token
     """
     if token is None:
-        raise HTTPException(403, 'Token missing from websocket connection query parameter')
+        raise WebSocketException(code=403, reason='Token missing from websocket connection query parameter')
```
krzysztof-causalens, line 417: I was under the impression that before we call `ws.accept` we haven't upgraded yet so we can still use `HTTPExceptions` here. TIL, thanks for catching that

krzysztof-causalens (approved):
LGTM, thanks!

PR #395, opened 2025-02-12
Fix: DO-5316 originating WebSocket channel being notified about their own variable updates

krzysztof-causalens:
## Motivation and Context

When using the `BackendStore` persistence for `Variable`s, the user who caused the update would also receive a WS message about the change. This would cause potential race conditions and state desyncs where if the variable changes multiple times while an update is being sent to the backend, the client can then get set 'back in time' after they receive a delayed notification about their own update.

This is especially noticeable on slower connections and when storing objects in the variables (as Recoil might skip updates for same primitive values).

## Implementation Description

Added `ws_channel` parameter to the request for `POST /api/core/store`. We then set the context var `WS_CHANNEL` before invoking `store.write()`.

Inside the backend store notification logic, we pass the current `WS_CHANNEL` into the websocket manager's `broadcast` and `send_message_to_user` methods as an `ignore_channel` kwarg.

Implemented the aforementioned `ignore_channel` kwarg to skip sending updates to a provided channel.

## How Has This Been Tested?

Verified in a local app that WS updates are sent to a different tab of the same user but not to themselves.

Updated client- and backend-side tests to cover the new behaviour. 

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/internal/routing.py
+++ b/packages/dara-core/dara/core/internal/routing.py
@@ -490,10 +490,11 @@ def create_router(config: Configuration):
         return result
 
     @core_api_router.post('/store', dependencies=[Depends(verify_session)])
-    async def sync_backend_store(values: Dict[str, Any] = Body()):
+    async def sync_backend_store(ws_channel: str = Body(), values: Dict[str, Any] = Body()):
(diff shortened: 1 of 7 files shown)


sam-causalens (approved):
LGTM, nice one 👍

PR #402, opened 2025-04-01
Fix: DO-5503 - Changed authentication of websockets

jsumiguin-causalens:
## Motivation and Context

Currently, websockets are authenticated using the tokens provided from the frontend. This change allows the user to override that authentication process depending on the dara auth config that they are using.

## Implementation Description

 - Changed authentication process to be via the `verify_token` method of the auth config instead of implemented by the web socket itself. This allows us to override how the authentication should work for different types of authentication configurations.
 - Used the `verify_token` functionality as that achieves the same thing.
 
N/A

## How Has This Been Tested?

Locally
## PR Checklist:


krzysztof-causalens (approved):
Looks good, have you tested it locally with all the built-in auth configs and our internal SSO one?

PR #403, opened 2025-04-07
Improvement: DO-5860 Allow dara to run behind a path prefixed proxy

sam-causalens:
## Motivation and Context
For users wishing to run a dara app behind a path prefixed proxy then this setting had to be exposed.

## Implementation Description
* Allow the pass through of a root path to the uvicorn ASGI server via param or env var to the dara start cli

No

## How Has This Been Tested?
Locally

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/cli.py
+++ b/packages/dara-core/dara/core/cli.py
@@ -60,6 +60,11 @@ def cli():
 @click.option('--log', default=lambda: os.environ.get('DARA_DEV_LOG_LEVEL', None), help='Dev logger level to use')
 @click.option('--reload-dir', multiple=True, help='Directories to watch for reload')
 @click.option('--skip-jsbuild', is_flag=True, help='Whether to skip building the JS assets')
+@click.option(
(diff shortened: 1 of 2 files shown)


packages/dara-core/dara/core/cli.py
```diff
@@ -76,6 +81,7 @@ def start(
     log: Optional[str],
     reload_dir: Optional[List[str]],
     skip_jsbuild: bool,
+    root_path: Optional[str],
```
krzysztof-causalens, line 84: Does this work with all the asset paths? We have another place adjusting the root path used for e.g. Jupyter integration where we pass `DARA_BASE_URL` already, to make images and other assets work. Have you tested if images work with this change?

sam-causalens: Ah, good catch I didn't know about this setting. I'll use that env var for the root path and make sure it's set to match if the flag is passed.

krzysztof-causalens (approved):

PR #404, opened 2025-04-07
Improvement: DO-5864 migrate to pydantic v2

krzysztof-causalens:
## Motivation and Context

Dara is currently stuck on pydantic v1 and thus so is the rest of our internal tooling. Ideally we'd have a transition period where we'd use pydantic's compatibility layer. 
Unfortunately this doesn't work with FastAPI since it automatically uses the v2 models if it detects pydantic v2 is available. 

Therefore this PR is a complete switch from v1 to v2.

## Implementation Description

Main changes:
- `@validator` -> `@field_validator`
- `class Config: ...` -> `model_config = ConfigDict(...)`
- `model.dict()` -> `model.model_dump()`
- REMOVED the template/For component feature due to its tight coupling to previous pydantic internals, it was barely used anyway
- Redid how serialization of models work due to pydantic's change to now only serialize properties of the declared type rather than the subclass. Achieved the old behaviour for our internals like components, actions, variables etc via a custom base model that modifies annotations to add the `SerializeAsAny` annotation to all types that are BaseModel classes. We also have to explicitly call `model_rebuild` which we override for all models to ensure that when the forward-declared types (due to our use of `from __future__ import annotations`) are made concrete we apply the serialization annotation. 

- bumped `fastapi-vite-dara` fork to a new release which uses pydantic.v1 compat mode internally
- bumped `pydantic` to v2.x
- added `pydantic-settings` since it was moved out of `pydantic` itself

## PR Checklist:


Roman-Kornev (approved):

sam-causalens (approved):
LGTM, thanks for this 🙏

krzysztof-causalens:
Note I noticed there's an issue with serializing custom types while playing with a demo:

```
  File "/Users/kbiel/code/dara/packages/dara-core/dara/core/visual/dynamic_component.py", line 339, in _render_safe
    normalized_data, normalized_lookup = normalize(jsonable_encoder(safe_result, custom_encoder=custom_encoder))
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kbiel/code/dara/.venv/lib/python3.11/site-packages/fastapi/encoders.py", line 223, in jsonable_encoder
    obj_dict = _model_dump(
               ^^^^^^^^^^^^
  File "/Users/kbiel/code/dara/.venv/lib/python3.11/site-packages/fastapi/_compat.py", line 177, in _model_dump
    return model.model_dump(mode=mode, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kbiel/code/dara/.venv/lib/python3.11/site-packages/pydantic/main.py", line 463, in model_dump
    return self.__pydantic_serializer__.to_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'cai_causal_graph.causal_graph.CausalGraph'>
```

I'm looking into how we can restore old behaviour and have known types be serialized automatically, but it seems like the old system doesn't work anymore

krzysztof-causalens:
Just to note the solution:
- values directly passed into `jsonable_encoder` still use our custom encoders (so e.g. returned from `DerivedVariable`)
- added `CausalGraph` and `Skeleton` into the default registry
- for serializing component properties, they need to be annotated on a **field** level with e.g. `@field_serializer`. Added a serializer for fields in `BaseGraphComponent` to handle graphs
- for serializing `Variable.default` field, added a `@field_serializer` that uses an appropriate encoder if possible and falls back to default serialization

Added a test for CausalGraphViewer with variable default to cover that scenario.

PR #417, opened 2025-04-22
Improvement: DO-6035 Enable custom js apps to have dynamic base urls

sam-causalens:
## Motivation and Context
The aim here is to allow dara apps with custom JS to be run behind a proxy.

## Implementation Description
* Adjust the vite config to push rendering of asset urls down to runtime
* Set the base_url via jinja templating rather than pushing it down through a function
* Remove the setting of baseUrl via the run function as it would get in the way of the template based setting
* Sets the static url correctly in the cli function to ensure that's correct
* Updates the injection of base_url into the autojs template to match so the two work in the same way.

No

## How Has This Been Tested?
Testing this locally with a studio build, will be pushing some dev tags to test things out after we ran into some build issues.

## PR Checklist:

## Screenshots

Jupyter running with new versions:
Screenshot 2025-04-24 at 13 56 40

Studio running behind a proxy locally with new versions:
Screenshot 2025-04-24 at 13 57 01


Roman-Kornev (approved):

krzysztof-causalens (approved):
LGTM, just missing a changelog entry

PR #422, opened 2025-05-20
Fix: DO-6306 Patch dara referrer handling when a base_url is set

sam-causalens:
## Motivation and Context
* The base_url wasn't being handled correctly when setting the referrer into state for redirection. This caused the path part of the base_url to be duplicated on redirect.

## Implementation Description
* Added a resolver function that adjusts the referrer based on the base_url.
* Adds a test to check it's working correctly.

No

## How Has This Been Tested?
Unit test added

## PR Checklist:

### DIFF
--- a/packages/dara-core/js/auth/auth.tsx
+++ b/packages/dara-core/js/auth/auth.tsx
@@ -121,6 +121,28 @@ export async function revokeSession(): Promise<RedirectResponse | SuccessRespons
 }
 
 /**
+ * Resolve the referrer url to be passed back to login, adjusted for the base path.
+ */
+export function resolve_referrer(): string {
+    if (!window.dara.base_url) {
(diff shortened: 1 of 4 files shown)


packages/dara-core/js/auth/auth.tsx
```diff
@@ -120,6 +120,28 @@ export async function revokeSession(): Promise<RedirectResponse | SuccessRespons
     }
 }
 
+/**
+ * Resolve the referrer url to be passed back to login, adjusted for the base path.
+ */
+export function resolve_referrer(): string {
```
krzysztof-causalens, line 126: Could we use camel case here?

sam-causalens: Oops, haven't written JS in a while 😅

krzysztof-causalens (approved):

PR #427, opened 2025-05-29
Improvement: DO-6248: Fix Tooltip infinite loop, Remove tippy.js, Migrate Tooltip to floating-ui

Roman-Kornev:
## Motivation and Context

Fixes an issue where clicking too fast in a controlled MultiSelect would cause an infinite loop. The cause was identified to be tippy.js. Since tippy.js is now deprecated and archived, we have migrated Tooltip to floating-ui, which we already use for other components. This was the last remaining tippy.js component, so we can now safely remove it from our deps.

## Implementation Description

The external API did not change and was reimplemented using floating-ui's hooks.

The only other usage was `GetReferenceClientRect` tippy.js type, which was replaced with a `DOMRect` implementation since it automatically calculates the other properties.

Removed tippy.js

## How Has This Been Tested?

Storybook and demo app. Added more stories.

## PR Checklist:

image
image

### DIFF
--- a/packages/ui-causal-graph-editor/package.json
+++ b/packages/ui-causal-graph-editor/package.json
@@ -105,7 +105,6 @@
     "react-dnd": "^14.0.5",
     "react-dnd-html5-backend": "^14.1.0",
     "svg-path-parser": "^1.1.0",
-    "tippy.js": "6.2.6",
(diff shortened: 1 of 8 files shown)


krzysztof-causalens (approved):
LGTM

PR #440, opened 2025-06-16
Feat: DO-6704 implement `run_task` method on ActionCtx

krzysztof-causalens:
## Motivation and Context

This PR adds a `run_task` method to `ActionCtx` to support imperatively executing tasks from actions.
Most of the functionality existed in the Dara internals already, we were only missing a way to handle progress.

## Implementation Description

Method added directly on the action ctx. Behind the scenes we create a `Task` object and submit that to the TaskManager, then await the resulting PendingTask and return/raise the result.

Added extra `on_progress` optional param to `Task`. In `TaskManager`, when provided, it's invoked on every task update received with the raw update message.

## How Has This Been Tested?

Locally, also checked that existing task system still works
Tests added

## PR Checklist:

https://github.com/user-attachments/assets/8176731c-f794-46c4-b7de-5885249ed0c4


Roman-Kornev (approved):

sam-causalens (approved):
Thanks Krys looks good. Can you add a section on this to the docs as well please?

PR #475, opened 2025-07-22
Docs: DO-7107: Sprinkle more raw_css docs

roman-causalens:
## Motivation and Context

The only page we have right now that mentions custom styling is the `docs/advanced/styling-components.md` page. It's better to sprinkle more mini-examples of how to style components throughout the docs for more complete examples.

## PR Checklist:


krzysztof-causalens (approved):
We should probably show both examples of dict-based styles and string styles.

Also separate point, technically raw_css takes `camelCase` properties (because React) but we turn kebab-case into camel under the hood so both are fine

PR #486, opened 2025-08-12
Fix: DO-7206 address table selection issue by providing a new api that can be passed and a flag to disable the old behaviour

jsumiguin-causalens:
## Motivation and Context

Currently, we run the click functionality whenever we select a row. This creates an unwanted behaviour of firing the click of the row when trying to just select it for multiselect purposes.

Also, fixed an issue where the urlvariable is not loaded on first load since the RecoilURLSync lives above the React Router.

## Implementation Description

 - Added two new props to the table -- `on_select_row` and `suppress_click_events_for_selection`.
 - Changed onSelect functionality to always fire onSelectRow, but don't fire onClickRow if it is suppressed.
 - Moved ReactRouter out of RecoilURLSync so that URLVariables will load correctly.
 
N/A

## How Has This Been Tested?

Locally

## PR Checklist:


packages/dara-components/js/common/table/table.tsx
```diff
@@ -626,17 +653,30 @@ function Table(props: TableProps): JSX.Element {
...
+            }
+
+            // If we don't want to suppress click events for selection, we want to trigger the click event as is before
+            if (!props.suppress_click_events_for_selection) {
+                onClickRow(selectedRows);
+            }
+
+            // If suppression is enabled, we want to trigger the click event and return the whole row
+            else if (!isCheckboxSelect) {
+                onClickRowSingle(cleanIndex([row]))
```
sam-causalens, line 669: I don't think you need the onClickRowSingle function? I'd imagine you just want to call onClickRow with the one row?

packages/dara-components/js/common/table/table.tsx
```diff
@@ -626,17 +653,30 @@ function Table(props: TableProps): JSX.Element {
                 }
 
                 setSelectedRowIndices(newSelectedIndices);
-                const selectedRows = await Promise.all(newSelectedIndices.map((idx) => getRowByIndex(idx)));
-                onClickRow(cleanIndex(selectedRows));
+                selectedRows = cleanIndex(await Promise.all(newSelectedIndices.map((idx) => getRowByIndex(idx))));
+            }
+            if (isCheckboxSelect && selectedRows !== null) {
```
sam-causalens, line 658: This `selectedRows !== null` should not be there otherwise you won't get an event when the user unselects the last row I think? If selectedRows is null then send []

packages/dara-core/changelog.md
```diff
@@ -2,6 +2,9 @@
 title: Changelog
 ---
 
+## 1.20.1-alpha.2
```
sam-causalens, line 5: Can you change this back to NEXT and put the UrlVariable bugfix here. The changelog entry for the table should be in the dara-components package not in core and if you could include and example that would be helpful.

sam-causalens (changes requested):
Looks good Jon, left a few comments to address before we can merge this.

sam-causalens (approved):
Thanks Jon, this looks better. Can you flip the changelog entries back to NEXT before you hit merge, but then happy for you to release this.

PR #488, opened 2025-08-19
Improvement: DO-7168 externalize pixi js library to lazy load from cdn

krzysztof-causalens:
## Motivation and Context

This PR removes pixi.js dependencies from the main Dara bundles, following the steps of bokeh.js and plotly libraries to ensure the largest dependencies which are often not required in apps aren't included in the main bundle. In the same vein, Pixi libraries are now lazily loaded from a CDN at the point of using the CausalGraphViewer component. This concludes the low-hanging fruit available for us to externalize.

This decreases the UMD size required as well as the prod bundle size for any dara app by ~20-25%

Before:

Screenshot 2025-08-19 at 11 53 27

Screenshot 2025-08-19 at 11 55 19

After:

Screenshot 2025-08-19 at 11 45 50

Screenshot 2025-08-19 at 11 58 43

## Implementation Description

Lazily loaded pixi.js, pixi-filters and pixi-viewport from `jsdelivr` and `unpkg` (couldn't find one CDN that had all 3 in the right format). 

Doing that uncovered an incompatibility between vitest and the MSW version we were using. Due to issues with bumping the MSW version and the existing jest setup in dara-core, I decided to upgrade to vitest to simplify the setup and avoid the issues altogether.

## How Has This Been Tested?

Existing tests passing
Checked the UMD and production builds work in a demo app

## PR Checklist:


packages/dara-components/package.json
```diff
@@ -77,7 +77,7 @@
     "@darajs/ui-utils": "1.20.1",
     "@popperjs/core": "2.4.0",
     "@vscode/codicons": "0.0.36",
-    "date-fns": "2.9.0",
+    "date-fns": "2.30.0",
```
krzysztof-causalens, line 80: When analyzing bundles I noticed date-fns was pulled in twice, so I upgraded to the higher version to eliminate the duplication

packages/ui-causal-graph-editor/src/pixi.d.ts
krzysztof-causalens, line 1: This file correctly aligns the type definitions to match how the CDN builds of the Pixi packages assign themselves - i.e. `pixi.js` becomes `PIXI` (and `pixi_js`), `pixi-filters` becomes `PIXI.filters` and `pixi_viewport` is a global `pixi_viewport`.

packages/ui-causal-graph-editor/package.json
```diff
@@ -58,14 +58,18 @@
     "graphology-generators": "^0.11.2",
     "jest": "^29.5.0",
     "jest-css-modules": "^2.1.0",
+    "pixi-filters": "~6.0.4",
+    "pixi-viewport": "5.0.3",
+    "pixi.js": "~8.5.0",
```
krzysztof-causalens, line 63: These are now type-only dependencies hence they're in `devDependencies`

PR #489, opened 2025-08-22
Feat: DO-7360 dara routing improvements - initial changes, backwards compatibility

krzysztof-causalens:
## Motivation and Context

This PR is the first step in the routing enhancements workstream. It introduces a new, more expressive and powerful `config.router` API.

I focused on ensuring backwards compatibility. Implemented:
- new Router API
- new components - Outlet, Link, MenuLink
- deprecated APIs intended for removal
    - `config.add_page`
    - `config.add_template_renderer`
    - `config.template` - replaced by the more powerful layout system, the built-in templates can easily be reproduced with a couple lines of code using SideBarFrame, MenuLinks and router APIs directly
    - `config.powered_by_causalens` - moved to be a SideBarFrame prop as well as a standalone component `PoweredByCausalens`

Not implemented yet:
- param variables
- guards & custom error pages
- optimized page data prefetching
- updates to documentation, changelogs

This should be a fine stage 1 where the existing APIs should function as before but before this PR gets too large with the new features above

## Implementation Description

Template compatibility strategy:
- on server startup, we check whether `config.router` or `config.pages` were used; if both, we raise an error to guide users towards migration
- if config.pages was used, we run a conversion function on the template to convert it to a new router
- we traverse the component tree, replacing the old RouterContent with Outlet, extracting routes from it and converting them to the new RouteDefinition structures

Page loading:
- each route has a ReactRouter client-loader defined which makes a singular request to the backend to fetch the template for that page
- this uses a tanstack-query suspense query, prefetching it in the loader; we use it over just a raw request call to enable us to implement link prefetching afterwards, as well as tune the caching if we decide we want to have a short-lived cache instead of making requests on every navigation 
- built-in link prefetching only works in framework in ReactRouter, so we'll implement it ourselves by simply pre-warming the react-query cache on e.g. link hover
- for now I kept a 5 min cache on page data, that will very likely change as I implement the next stage where we optimize prefetching page data and running the on_load action within the same request, as well as implement page guards

Other issues fixed/optimizations made:
- the initial HTML is no longer blank and includes a full-page loader while assets are loading
- unified template handling to use jinja for both branches, previously the autojs template used manual string replacement
- autojs assets now use a query param `?v={version}` to prevent stale browser cache when changing versions
- components are now preloaded based on initial registry which reduces flashes of spinners on the first time components are used on the page
- removed a waterfall of 3-4 requests happening on initial page loads by embedding the data directly in the HTML in a script `id="__DARA_DATA__"`

## PR Checklist:

Navigation indicators using NProgress - note for now route requests are cached for 5 mins, that will change as we implement page guards and optimize prefetching page data

https://github.com/user-attachments/assets/86cc3672-fecc-4187-a86c-11386c61133f

Loading indicator immediately displayed while assets are loading

https://github.com/user-attachments/assets/2783f1f2-9ca2-4bd3-8cd5-0c223d0c3ad3


packages/dara-core/dara/core/main.py
```diff
@@ -72,6 +74,16 @@
     rebuild_js,
 )
 from dara.core.logging import LoggingMiddleware, dev_logger, eng_logger, http_logger
+from dara.core.router import convert_template_to_router
+
+
+class CacheStaticFiles(StaticFiles):
+    async def get_response(self, path, scope):
+        response = await super().get_response(path, scope)
+        # add 1 year cache for static assets
```
krzysztof-causalens, line 83: We can do that safely now that we know the file names are unique (thanks to the `?v=` suffixes for autojs assets, and vite assets are hashed anyway)

packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
```diff
@@ -90,6 +94,63 @@ export function clearCaches_TEST(): void {
...
+            throw new Error(`Failed to load module ${pyModule}: ${String(e)}`);
+        }
+
+        // pre-load components
+        for (const component of componentsInModule) {
+            if (COMPONENT_METADATA_CACHE.has(component.name)) {
+                continue;
+            }
+            COMPONENT_METADATA_CACHE.set(component.name, component);
+        }
```
Roman-Kornev, line 150: What happens if components have the same name but different content?

packages/dara-core/js/shared/root/authenticated-root.tsx
```diff
@@ -0,0 +1,132 @@
...
+    useEffect(() => {
+        // already set up - make sure we don't recreate connections,
+        // we use updateToken explicitly to update a live connection
+        if (wsClient) {
+            return;
+        }
+
+        const newWsClient = setupWebsocket(token, props.daraData.live_reload);
+        setWsClient(newWsClient);
+    }, [token, props.daraData.live_reload]);
```
Roman-Kornev, line 103: Effect feels weird without `wsClient` in deps, but then it uses `setWsClient`. Surprised the linter doesn't complain.

packages/dara-core/js/shared/router.tsx
```diff
@@ -0,0 +1,242 @@
...
+    return path.replace(/^\/+|\/+$/g, '');
+}
+
+/**
+ * Helper function to clean and join path segments properly
+ */
+function joinPaths(parentPath: string, childPath: string): string {
+    // Remove leading/trailing slashes from both parts
+    const cleanParent = parentPath.replace(/^\/+|\/+$/g, '');
+    const cleanChild = childPath.replace(/^\/+|\/+$/g, '');
```
Roman-Kornev, line 35: Reuse cleanPath?

packages/dara-core/js/shared/router.tsx
```diff
@@ -0,0 +1,242 @@
...
+        // 1. Prefer index routes - return current path as it's navigatable
+        if (indexRoute) {
+            return currentPath === '' ? '/' : joinPaths('', currentPath);
+        }
+
+        // 2. Then prefer page routes - just return them joined with parent
+        for (const pageRoute of pageRoutes) {
+            const pagePath = joinPaths(currentPath, pageRoute.path);
+            return pagePath;
+        }
```
Roman-Kornev, line 128: Both can early return in the `currentRoutes` matching stage, no need to store them

Roman-Kornev (approved):

packages/dara-core/dara/core/jinja/index.html
```diff
@@ -6,6 +6,98 @@
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <link id="favicon" rel="icon" type="image/x-icon" href="{{ static_url }}/favicon.ico" />
 
+        <!-- inline style block to show a spinner before js & css loads -->
```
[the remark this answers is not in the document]
krzysztof-causalens: Well yeah but that's better than an infinite white page 😄

packages/dara-core/dara/core/js_tooling/js_utils.py
```diff
@@ -619,27 +626,31 @@ def prepare_autojs_assets(build_cache: BuildCache):
...
+        # Note: this assumes assets structure of module/umd folder with the module.umd.(c)js and style.css file
         js_asset_path = os.path.join(module_path, 'umd', f'{module_name}.umd.js')
+        cjs_asset_path = os.path.join(module_path, 'umd', f'{module_name}.umd.cjs')
         css_asset_path = os.path.join(module_path, 'umd', 'style.css')
 
         # copy over the JSS/CSS from python package into dist/umd so they are available under /static
         if os.path.exists(js_asset_path):
             shutil.copy(js_asset_path, os.path.join(build_cache.static_files_dir, f'{module_name}.umd.js'))
+        elif os.path.exists(cjs_asset_path):
+            shutil.copy(cjs_asset_path, os.path.join(build_cache.static_files_dir, f'{module_name}.umd.js'))
```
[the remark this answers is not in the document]
krzysztof-causalens: No, we still copy it to `dist` as `.js`, it's just that the source file (i.e. the assets stored in the wheels) can be `.cjs` now.

packages/dara-core/dara/core/router.py
```diff
@@ -0,0 +1,884 @@
...
+
+
+class HasChildRoutes(BaseModel):
+    children: SerializeAsAny[List['BaseRoute']] = Field(default_factory=list)
+
+    def __init__(self, *children: list[BaseRoute], **kwargs):
+        routes = list(children)
+        if 'children' not in kwargs:
+            kwargs['children'] = routes
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
krzysztof-causalens: This will just error that you shouldn't pass both, I don't think we need to support passing both at the same time - it's just convenience. This behaviour is consistent with how our components work

packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
krzysztof-causalens: Dara has always assumed component names are unique, this isn't a new assumption. We haven't really ran into issues with it so far, but I would be open to revisiting this in the future

packages/dara-core/js/shared/root/authenticated-root.tsx
krzysztof-causalens: True, it probably should have it since we have the explicit wsClient check here

PR #492, opened 2025-09-02
Fix: DO-7444 - Fix click propagation from buttons so they can be used inside other buttons

sam-causalens:
## Motivation and Context
Fix an issue for using buttons in buttons

## Implementation Description
* Adds a flag that allows the old behaviour to be restored if needed

No

## How Has This Been Tested?
Tested Locally

## PR Checklist:

### DIFF
--- a/packages/dara-components/tests/python/common/test_button.py
+++ b/packages/dara-components/tests/python/common/test_button.py
@@ -24,6 +24,7 @@ class TestButtonComponent(unittest.TestCase):
             'props': {
                 'bold': False,
                 'children': [Text(text='Click Here').dict(exclude_none=True)],
+                'stop_click_propagation': True,
                 'italic': False,
                 'onclick': action.dict(),
                 'styling': ButtonStyle.PRIMARY.value,


packages/dara-components/js/common/button/button.tsx
```diff
@@ -77,7 +79,12 @@ function Button(
             disabled={disabledValue}
             isSimpleButton={typeof children === 'string' || children[0]?.name === 'Text'}
             loading={loading}
-            onClick={() => onClick(null)}
+            onClick={(e: React.MouseEvent<HTMLButtonElement>) => {
+                if (props.propagate_click_event !== false) {
```
krzysztof-causalens, line 83: I think the naming is reverted? It should be `if (propagate === false) stopPropagation()` .

Maybe let's just rename to e.g. `stop_click_propagation: bool = True`

krzysztof-causalens (approved):
Looks good once the naming is fixed, thanks!

packages/dara-components/js/common/button/button.tsx
sam-causalens: Yep, good point 👍 Fixed now.

PR #493, opened 2025-09-03
Improvement: DO-7442 dara routing changes - documentation updates, DX improvements

krzysztof-causalens:
## Motivation and Context

This PR updates all documentation to use the new routing APIs and adds a new `Routing` docs page to replace the old "custom templates" page. 

In addition, a few DX improvements were made to make transitioning from the old API smoother:
- `content=` kwarg on `add_page` et al now accepts a component in addition to a function; when param injections are not necessary this avoids users having to wrap components in empty lambdas, we can do it for them
- `config.router` is now populated with an empty `Router` by default; this makes the change from the old API easier (e.g. for simple cases just change `config.add_page` to `config.router.add_page` and update kwarg names). Users are still free to create a new router instance if they wish

## How Has This Been Tested?

Manually, added few tests for the changed features

## PR Checklist:


docs/quickstart.md
```diff
@@ -103,21 +103,62 @@ from dara.components import Heading
...
 ```sh
 poetry run dara start
 ```
 
-Your app will look like the following:
+Your app will be a blank page with the text `Hello World!` on it.
+
+### Menu
+
+Your are off to a good start but we can do better. As we add more pages, we will want the user to be able to navigate to them easily - currently, the only way to do this would be to navigate to the right URL manually.
```
Roman-Kornev, line 121:
```suggestion
You are off to a good start but we can do better. As we add more pages, we will want the user to be able to navigate to them easily - currently, the only way to do this would be to navigate to the right URL manually.
```

docs/quickstart.md
```diff
@@ -103,21 +103,62 @@ from dara.components import Heading
...
+            MenuLink(Text('Hello World'), to='/hello-world'),
+        )
+    )
+
+root = config.router.add_layout(content=Layout)
+root.add_page(path='hello-world', content=Heading('Hello World!'))
+```
+
+Notice how we are using `router.add_layout` to wrap the entire application in a layout built using prebuilt sidebar and menu link components.
+The returned value of the function is a `LayoutRoute` - rather than adding our "hello-world" page to the router directly, we now nest it under that layout.
```
Roman-Kornev, line 147: Worth explaining what an `Outlet()` is and that the Heading will replace it in the component tree

Roman-Kornev (approved):

PR #495, opened 2025-09-05
Improvement: DO-7496 - propagate the `id_` field to the DOM

momustafa-causalens:
## Motivation and Context
When using the pythonic dara components, we can set the `id_` property, but this is not propagated to the frontend. This change propagates the property to the DOM.

## Implementation Description
- Introduced `id` prop to multiple components including Accordion, Button, Card, and more.
- Updated existing components to utilize `props.id_` for setting the `id` attribute in the rendered output.

## PR Checklist:

### DIFF
--- a/packages/dara-components/js/plotting/bokeh/bokeh.tsx
+++ b/packages/dara-components/js/plotting/bokeh/bokeh.tsx
@@ -61,7 +61,8 @@ function Bokeh(props: BokehProps): JSX.Element {
 
     const docJson = useMemo<DocJson>(() => JSON.parse(props.document), [props.document]);
     const rootId = useMemo(() => docJson.roots[0]!.id, [docJson]);
-    const id = props.id_ ?? useId();
+    const uid = useId();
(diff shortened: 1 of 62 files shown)


packages/dara-components/js/plotting/bokeh/bokeh.tsx
```diff
@@ -61,7 +61,7 @@ function Bokeh(props: BokehProps): JSX.Element {
 
     const docJson = useMemo<DocJson>(() => JSON.parse(props.document), [props.document]);
     const rootId = useMemo(() => docJson.roots[0]!.id, [docJson]);
-    const id = useId();
+    const id = props.id_ ?? useId();
```
krzysztof-causalens, line 64: Conditionally calling hooks is dangerous, should be fine but better to do

```suggestion
    const uid = useId();
    const id = props.id_ ?? uid;
```

krzysztof-causalens (approved):
Spotted a minor issue but LGTM otherwise, thanks

PR #624, opened 2026-04-20
Feat: DO-10894 - Added LocalizedDate Component

momustafa-causalens:
## Motivation and Context
We want a component that renders an ISO date string formatted in the user's local timezone, using date-fns format tokens.

## Implementation Description
Added that component

No

## How Has This Been Tested?
Manually

## PR Checklist:

### DIFF
--- a/packages/dara-components/dara/components/common/localized_date.py
+++ b/packages/dara-components/dara/components/common/localized_date.py
@@ -15,8 +15,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
-from pydantic import field_validator
-
 from dara.components.common.base_component import ContentComponent
 from dara.core.interactivity import ClientVariable
 
@@ -25,6 +23,10 @@ class LocalizedDate(ContentComponent):
     """
     A LocalizedDate component renders an ISO date string formatted in the user's
     local timezone. Formatting follows [date-fns](https://date-fns.org/docs/format) tokens.
+
+    When `date` is empty or cannot be parsed, `placeholder` is rendered instead and
+    `data-state="error"` is set on the wrapper so the error state can be targeted via
+    `raw_css`.
 
     A LocalizedDate component can be created like so:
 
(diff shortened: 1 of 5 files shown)


packages/dara-components/dara/components/common/localized_date.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    ```
+
+    :param date: The date to display, as an ISO string or a Variable resolving to one
+    :param format: The date-fns format string, defaults to 'yyyy-MM-dd HH:mm'
+    """
+
+    date: str | ClientVariable
+    format: str = 'yyyy-MM-dd HH:mm'
```
krzysztof-causalens: Might as well support a variable here too

packages/dara-components/js/common/localized-date/localized-date.tsx
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param props the component props
+ */
+function LocalizedDate(props: LocalizedDateProps): JSX.Element {
+    const { date, format: formatString = 'yyyy-MM-dd HH:mm' } = props;
+    const [style, css] = useComponentStyles(props);
+    const [dateStr] = useVariable(date);
+
+    const parsedDate = parseISO(dateStr);
+    const formattedDate = format(parsedDate, formatString);
```
krzysztof-causalens: We should probably handle invalid dates here, maybe user can provide a `fallback` to the component? I guess `fallback` is an overloaded term so maybe `default` or something instead. 

Ideally we also let the user style the errored date as well, so maybe we apply some data property so user can do `raw_css="&[data-state="error"] { color: red; }"` if they want to

packages/dara-components/dara/components/common/localized_date.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @field_validator('date')
+    @classmethod
+    def validate_date(cls, date):
+        if not isinstance(date, str | ClientVariable):
+            raise ValueError('Date must be a string or Variable')
+        return date
```
krzysztof-causalens: This should be checked by Pydantic already, redundant

krzysztof-causalens (approved):

PR #626, opened 2026-04-28
Fix: DO-10971 - preserve header columns and stabilize filter

momustafa-causalens:
## Motivation and Context

When a text filter is applied to a Table column, the filter input popover closes on its own ~500ms after typing a character. Additionally, if the filter yields zero results, the entire column header row disappears.

The root cause is that `fetchData` calls `setResolvedColumns` with new objects on every fetch — even when the columns haven't changed. This triggers a cascade: new columns → react-table reconstructs header groups → `FilterContainer` remounts → `isOpen` state resets to `false`.

For the empty-data case, `setResolvedColumns([])` was explicitly wiping the columns, removing the header entirely.

## Implementation Description

Two targeted changes in `packages/dara-components/js/common/table/table.tsx`:

- **Stabilize `setResolvedColumns`**: use a functional updater with `isEqual` (already imported) to preserve the existing reference when columns haven't semantically changed. This prevents the re-render cascade that unmounts the filter popover.
- **Remove `setResolvedColumns([])` on empty data**: when a fetch returns no rows, simply return early without wiping columns. The header stays visible with the previously resolved columns.

None.

## How Has This Been Tested?

- Apply a text filter to a table column — verify the filter input stays open while typing
- Filter to a value that yields zero results — verify the column header row remains visible
- Verify sorting and other filter types still work as expected after the change

## PR Checklist:

### BEFORE

https://github.com/user-attachments/assets/9740906e-ec0f-424c-9f28-f8527a291866

### AFTER

https://github.com/user-attachments/assets/56436f4f-d7f2-4c28-8ac7-aa8b4abafd08

### DIFF
--- a/packages/dara-components/changelog.md
+++ b/packages/dara-components/changelog.md
@@ -6,11 +6,14 @@ title: Changelog
 
 - Fixed table column text filter closing immediately after typing by stabilizing resolved columns reference
 - Fixed table header row disappearing when a column filter returns no results
+- Fixed linting issues: `Accordion.validate_initial` return type, `CustomJS` import path, and `Plotly` figure optional access
 
 ## 1.26.8
+
 - Added `LocalizedDate` component to the `dara.components.common` package
 
 ## 1.26.5
+
 - Removed unused sortType code for the Table's Columns, here since the sorting is handled in different places for dara-components
 
 ## 1.26.2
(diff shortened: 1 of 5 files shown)


packages/dara-components/changelog.md
```diff
@@ -2,10 +2,18 @@
 title: Changelog
 ---
 
+## NEXT
+
+- Fixed table column text filter closing immediately after typing by stabilizing resolved columns reference
+- Fixed table header row disappearing when a column filter returns no results
+- Fixed linting issues: `Accordion.validate_initial` return type, `CustomJS` import path, and `Plotly` figure optional access
```
krzysztof-causalens: not relevant to end users

krzysztof-causalens (approved):
Looks good, thanks! Just one small note about the changelog

krzysztof-causalens:
@codex review

PR #627, opened 2026-04-28
Fix: DO-11004 - Set resolvedColumns to empty array on first load so table exits loading state

momustafa-causalens:
## Motivation and Context
If the first fetchData call returns an empty result (for example an initially empty dataframe or a startup filter that matches nothing), this branch now exits without updating resolvedColumns. The render path only shows <UiTable> when resolvedColumns is truthy and otherwise shows <DefaultFallback />, so the table gets stuck in a perpetual loading state instead of rendering an empty table. This is a regression from the previous behavior that set resolvedColumns on empty results.

## Implementation Description
Set resolvedColumns to empty array on first load so table exits loading state

No

## How Has This Been Tested?
Manually in demo app

## PR Checklist:


krzysztof-causalens:
@codex review

krzysztof-causalens (approved):
Missing changelog entry

PR #630, opened 2026-05-12
Fix: DO-11207 Adjust the adaptive formatter for small numbers to show more decimal places

sam-causalens:
## Motivation and Context
Tweak to the adapative cell formatter for small numbers.

## Implementation Description
* Show 4dp when numbers are between 0.1 and 0.001 show 4 numbers rather than only 3. Workloads in this range can often need the extra precision and an extra 0 won't look odd given the range of the numbers

No

## How Has This Been Tested?
Locally

## PR Checklist:


krzysztof-causalens (approved):
Thanks, could you add a changelog entry?

PR #632, opened 2026-05-20
Fix: DO-11363 missing auth cookies redirect to login, only refresh token on 401 not 403 status

krzysztof-causalens:
## Motivation and Context

When a browser session cookie expires and is no longer sent, Dara can receive authenticated API requests with no bearer token and no session cookie. The backend classified that as a `400 bad_request`, which the frontend routes to `/error?code=400` rather than back to login.

The client also attempted refresh-token recovery for `403` responses. A `403` means the user is authenticated but not authorized, so refreshing the session is not expected to resolve it and can mask the authorization failure.

## Implementation Description

Missing auth credentials now return the existing expired-session auth payload with HTTP `401`, so the frontend treats the response as an authentication failure and redirects users to login with the correct session-expired reason. Malformed `Authorization` headers still return `400`, preserving the bad-request behavior for invalid auth syntax.

The HTTP client now attempts `/api/auth/refresh-token` only for `401` responses. `403` responses are returned to callers for normal auth-error handling without using the refresh token.

Backend auth tests cover the missing-credentials path on `/verify-session`, update the general protected-endpoint expectation, and keep stale-cookie cleanup behavior aligned with the new status. JS HTTP tests now cover refresh-on-401 and no-refresh-on-403.

No new dependencies.

## How Has This Been Tested?

- `cd packages/dara-core && DARA_TEST_FLAG=True poetry run pytest tests/python/test_auth.py -q`
- `cd packages/dara-core && pnpm exec vitest run tests/js/http.spec.tsx --reporter=verbose`
- `poetry anthology run lint`
- `pnpm lerna run lint`
- `pnpm lerna run format:check`
- `git diff --check`

## PR Checklist:

- [x] I have added comments to all the bits that are hard to follow. N/A: no non-obvious code added.
- [x] I have added/updated Documentation. N/A: behavior fix covered by changelog.

N/A

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+- Fixed missing auth session cookies being reported as bad requests instead of unauthorized session failures, so expired browser cookies redirect users back to login, and limited refresh-token retries to authentication failures.
+
 ## 1.26.12
 
 - Internal: Hardened JavaScript dependency versions and removed unused build and test tooling dependencies.
--- a/packages/dara-core/dara/core/auth/routes.py
+++ b/packages/dara-core/dara/core/auth/routes.py
@@ -170,7 +170,7 @@ def _get_auth_token(
     if session_cookie_token is not None:
         return session_cookie_token
 
-    raise HTTPException(status_code=400, detail=BAD_REQUEST_ERROR(missing_message))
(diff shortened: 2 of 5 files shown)


krzysztof-causalens:
@codex review

krzysztof-causalens:
@codex review

sam-causalens (approved):
LGTM

PR #640, opened 2026-06-04
Fix: DO-11593 preserve login referrer encoding

krzysztof-causalens:
## Motivation and Context

Unauthenticated first loads in OIDC apps could hit `/error?code=400` before the identity provider flow started. The root auth redirect passed an already encoded referrer through `URLSearchParams`, double-encoding values such as `/` into `%252F`. `/login` decoded that once and sent `%2F` as `redirect_to`, which the backend correctly rejected as an invalid redirect target.

The first fix also surfaced a related client-side inconsistency: login components were manually calling `decodeURIComponent()` after `URLSearchParams.get('referrer')`, which already decodes once. That could turn an intended target such as `/files/foo%2Fbar?x=a%2Fb` into `/files/foo/bar?x=a/b` after login.

## Implementation Description

The root router login redirect now preserves the encoded referrer string when building `/login?referrer=...`, avoiding the extra `URLSearchParams` encoding pass.

Referrer handling is now split consistently by boundary:

- encoded referrer strings are used when constructing login URLs
- parsed referrer paths are used internally for navigation and OIDC `redirect_to` request bodies

Basic, default, and OIDC login flows now use the shared parsed referrer helper instead of decoding independently. Auth-error handling also re-encodes an existing login referrer before recycling it into a fresh `/login` redirect.

Updated the `dara-core` changelog with a high-level note for the fix.

None.

## How Has This Been Tested?

- `pnpm --filter @darajs/core exec vitest --run tests/js/auth.spec.tsx tests/js/router.spec.tsx tests/js/websocket-auth-recovery.spec.tsx tests/js/stream-variable.spec.tsx`
- `pnpm --filter @darajs/core exec tsc --noEmit`
- `git diff --check`

Tested locally - before I could reproduce that loading an OIDC-secured page without a cookie would always trigger 400 and redirect to `/error?code=400` instead of `/login`. After the fix user lands on `/login` as expected.

## PR Checklist:

- [x] I have added/updated the changelog for the relevant package.

N/A

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+- Fixed auth login redirects so unauthenticated first loads and referrers containing encoded path/query characters are preserved correctly.
+
 ## 1.28.0
 
 - Internal: Upgraded FastAPI to the latest 0.136 release line and explicitly require Starlette 1.0.1+ for the BadHost CVE fix.
--- a/packages/dara-core/js/auth/auth.tsx
+++ b/packages/dara-core/js/auth/auth.tsx
@@ -125,7 +125,7 @@ export async function revokeSession(): Promise<RedirectResponse | SuccessRespons
 }
 
 /**
- * Resolve the referrer url to be passed back to login, adjusted for the base path.
+ * Resolve the encoded referrer url to be passed back to login, adjusted for the base path.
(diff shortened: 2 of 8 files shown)


krzysztof-causalens:
@codex review

krzysztof-causalens:
@codex review

zack-causalens (approved):

PR #641, opened 2026-06-05
Fix: DO-11607 align OIDC ID token signing algorithm policy

krzysztof-causalens:
## Motivation and Context

The OIDC ID token verifier was selecting its accepted signing algorithm purely from the legacy `SSO_JWT_ALGO` setting, with an ES256 default. That did not line up with OIDC provider discovery and made the verification policy too detached from the configured issuer metadata.

OpenID Connect Discovery 1.0 Section 3 defines `id_token_signing_alg_values_supported` as provider capability metadata and says `RS256` must be included. Our internal IDP currently advertises only `ES256`, so the implementation uses the provider's discovered JWKS-verifiable non-`none` signing algorithms when no env override is configured, rather than rejecting startup solely because `RS256` is absent.

Explicit env configuration still represents Dara's local verifier policy and remains exact. This avoids treating the JWT `alg` header as verification policy. RFC 8725 Section 3.1 requires callers to perform algorithm verification, so the verifier now checks the token's claimed algorithm against the startup-resolved allow-list before decoding. If the selected JWK explicitly declares an `alg`, that key-level constraint must also match the token header.

## Implementation Description

Added `SSO_ID_TOKEN_SIGNED_RESPONSE_ALG` as the preferred env var and kept `SSO_JWT_ALGO` as a deprecated compatibility alias. Both settings default to `None`; explicit precedence is new env var, then legacy env var. Constructor compatibility for `jwt_algo` remains, but unprefixed env names like `JWT_ALGO` are not accepted.

If neither env var is configured, OIDC startup uses the JWKS-verifiable non-`none` algorithms advertised by `id_token_signing_alg_values_supported`, falling back to `RS256` only if discovery does not provide a usable signed algorithm. The discovery model accepts an omitted `id_token_signing_alg_values_supported` field as a compatibility concession so that fallback is reachable for non-compliant IDPs.

OIDC startup resolves and registers the allowed ID token signing algorithm list next to the JWKS client. Discovery is used as provider metadata, with `none` and HMAC algorithms filtered out because Dara verifies signed ID tokens using issuer JWKS. OpenID Connect Core 1.0 Section 10.1 uses client-secret-based HMAC for MAC-based signatures, which this verifier does not implement.

ID token verification now treats the JWT header `alg` as an untrusted claim of which cryptographic operation to perform. It is accepted only if it is in the startup-resolved env/discovery policy. The selected JWK's optional `alg` member is enforced when present, but PyJWT's key-type default is not used as policy when the OP omits `alg` from JWKS, preserving providers that use allowed RSA algorithms such as `PS256` with bare RSA JWKs.

Updated OIDC docs, changelog, and backend tests for the new precedence/default/discovery behavior, ES256 verification, malformed discovery fallback, algorithm/key mismatch rejection, and RSA JWKs that omit `alg`.

None.

## How Has This Been Tested?

- `cd packages/dara-core && poetry run ruff format dara/core/auth/oidc/settings.py dara/core/auth/oidc/config.py dara/core/auth/oidc/utils.py dara/core/auth/oidc/definitions.py tests/python/test_oidc_auth.py tests/python/test_oidc_settings.py`
- `cd packages/dara-core && poetry run ruff check --select I dara/core/auth/oidc/settings.py dara/core/auth/oidc/config.py dara/core/auth/oidc/utils.py dara/core/auth/oidc/definitions.py tests/python/test_oidc_auth.py tests/python/test_oidc_settings.py`
- `cd packages/dara-core && DARA_TEST_FLAG=True poetry run pytest tests/python/test_oidc_settings.py tests/python/test_oidc_auth.py`
- `poetry anthology run lint`
- `git diff --check -- packages/dara-core/changelog.md packages/dara-core/dara/core/auth/oidc/config.py packages/dara-core/dara/core/auth/oidc/definitions.py packages/dara-core/dara/core/auth/oidc/settings.py packages/dara-core/dara/core/auth/oidc/utils.py packages/dara-core/docs/advanced/authentication.mdx packages/dara-core/tests/python/test_oidc_auth.py packages/dara-core/tests/python/test_oidc_settings.py`

## PR Checklist:

- [x] I have added/updated the changelog for the relevant package.

N/A

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -4,7 +4,7 @@ title: Changelog
 
 ## NEXT
 
-- Changed OIDC ID token signing algorithm handling to support `SSO_ID_TOKEN_SIGNED_RESPONSE_ALG`, use JWKS-verifiable non-`none` discovery algorithms when no explicit algorithm is configured, and fall back to RS256.
+- Fixed OIDC ID token signing algorithm handling to match the spec better while maintaining escape hatches for non-compliant IDPs. The `SSO_ID_TOKEN_SIGNED_RESPONSE_ALG` env var remains as an override; JWKS-verifiable non-`none` discovery algorithms are preferred when no explicit algorithm is configured, with a fallback to RS256.
 - Fixed auth login redirects so unauthenticated first loads and referrers containing encoded path/query characters are preserved correctly.
 
 ## 1.28.0
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -4,6 +4,7 @@ title: Changelog
 
 ## NEXT
 
+- Changed OIDC ID token signing algorithm handling to support `SSO_ID_TOKEN_SIGNED_RESPONSE_ALG`, use JWKS-verifiable non-`none` discovery algorithms when no explicit algorithm is configured, and fall back to RS256.
 - Fixed auth login redirects so unauthenticated first loads and referrers containing encoded path/query characters are preserved correctly.
 
 ## 1.28.0
--- a/packages/dara-core/dara/core/auth/oidc/config.py
+++ b/packages/dara-core/dara/core/auth/oidc/config.py
@@ -37,13 +37,14 @@ from ..definitions import (
 )
 from ..utils import decode_token, sign_jwt
 from .definitions import (
+    ID_TOKEN_SIGNING_ALGS_REGISTRY_KEY,
     JWK_CLIENT_REGISTRY_KEY,
     IdTokenClaims,
     OIDCDiscoveryMetadata,
     OIDCLoginTransaction,
 )
 from .routes import sso_callback
-from .settings import OIDCSettings, get_oidc_settings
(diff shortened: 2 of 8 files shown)


krzysztof-causalens:
@codex review

krzysztof-causalens:
Addressed the RSA/JWK `alg` omission case in 2a3a2433. The verifier now treats the JWT header `alg` as an untrusted claimed operation, checks it against the startup-resolved env/discovery allow-list, and only enforces JWK `alg` when the OP explicitly provides it. If the JWK omits `alg`, we no longer use PyJWT's key-type default as policy, so allowed RSA algorithms like `PS256` work with bare RSA JWKs. Added regression coverage for both the omitted-`alg` case and explicit JWK/header mismatch.

krzysztof-causalens:
@codex review

muad-causalens (approved):
Not overly familiar with the Dara packages so not too sure if the code has been structured correctly and placed in the right places. The implementation itself LGTM

sam-causalens (approved):

PR #643, opened 2026-06-11
Fix: DO-11692 - abort stale StreamVariable SSE connections on dependency change

momustafa-causalens:
Fix: DO-11692 Abort stale StreamVariable SSE connections on dependency change

## Motivation and Context

When a `StreamVariable`'s dependencies change (e.g. page number in a paginated stream), the client opens a new SSE connection for the updated values but does not close the old one. The old server-side generator keeps running to completion, causing:

- Orphaned SSE connections accumulating with each dependency change
- Stale data being fetched and yielded before new data arrives
- Compounding latency (observed ~1s first change degrading to ~8s+ by the third change)

The root cause is in the stream usage tracker: subscriptions are keyed by `uid+extras`, but connections are keyed by `atomKey` (which includes resolved dependency values). When deps change, a new `atomKey` is produced for the same `uid+extras`, so `registerStreamConnection` starts a new connection without aborting the old one. The old connection is only cleaned up by the 2-minute orphan timer or unreliable Recoil atom release.

## Implementation Description

In `registerStreamConnection`, before starting a new connection, iterate the existing connections for the same subscription key (`uid+extras`) and abort + delete any whose `atomKey` differs from the one being registered. This enforces the invariant: at most one active SSE connection per stream variable per auth context.

The fix is client-side only. The server-side `is_disconnected()` check at yield points is sufficient -- once the client aborts the old `fetch`, the server detects the disconnect at the next yield and exits the generator cleanly.

Stale entries are deleted from the connections map (not just aborted) to prevent `restartConnections` from resurrecting dead connections on re-subscribe.

None. `pnpm-lock.yaml` updated from dependency resolution changes.

## How Has This Been Tested?

- 3 new unit tests in `use-stream-variable.spec.tsx`:
  - Single dependency change aborts old connection, keeps new one active
  - Different stream uids don't interfere with each other
  - Rapid sequential changes (page 1 -> 2 -> 3) abort each prior connection correctly
- All 317 existing tests pass with no regressions (30 test files)

## PR Checklist:

N/A - backend/client-side logic change, no UI changes.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -410,6 +410,24 @@ importers:
         specifier: ^3.6.20
         version: 3.6.20
 
+  packages/demo-app/dist:
+    dependencies:
+      '@darajs/components':
+        specifier: 1.26.8
+        version: 1.26.8(@egjs/hammerjs@2.0.17)(@lezer/common@1.2.3)(@types/hoist-non-react-statics@3.3.5)(@types/node@18.19.64)(@types/react@18.2.60)(component-emitter@1.3.1)(keycharm@0.4.0)(react-is@18.3.1)(uuid@11.1.0)(vis-util@5.0.7(@egjs/hammerjs@2.0.17)(component-emitter@1.3.1))
+      '@darajs/core':
+        specifier: 1.26.8
+        version: 1.26.8(@types/react@18.2.60)(react-is@18.3.1)
+      '@vitejs/plugin-react':
+        specifier: 4.6.0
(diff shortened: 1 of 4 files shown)


pnpm-lock.yaml
krzysztof-causalens, line 1: Shouldn't need any changes to the lockfile, no deps were changed

krzysztof-causalens (approved):
Good catch! Could you just remove the unnecessary pnpm-lock changes pls

PR #645, opened 2026-06-12
Fix: DO-11694 - cancel StreamVariable inner connections on client disconnect

momustafa-causalens:
Fix: StreamVariable inner connection leak on client disconnect (DO-11694)

## Motivation and Context

PR #643 fixed the client-side issue — when a `StreamVariable`'s dependencies change, the JS runtime now aborts the old `EventSource` before opening a new one. This correctly closes the **outer** SSE connection (browser → Dara server).

However, there is a second connection involved. A `StreamVariable` generator may open its own long-lived HTTP connections internally (e.g. connecting to an upstream SSE endpoint). When the client aborts the outer connection, the server-side generator remains suspended in an `await` on that **inner** HTTP connection — it is never cancelled.

```
Browser  ──(outer SSE)──▶  Dara server  ──(inner SSE)──▶  API server
   │                            │                              │
   │  Client aborts this ✓     │  Generator stuck here ✗      │
   │                            │  async for event in          │
   │                            │    api.stream_events()       │
```

Each stream restart (e.g. pagination) leaves an orphaned inner connection on the server. After a few page changes, accumulated connections exhaust the browser's per-origin limit (6 for HTTP/1.1), causing all subsequent requests to queue as "Pending".

## Implementation Description

The root cause is in `run_stream` (`stream_variable.py`). The previous loop:

```python
async for event in generator:
    if await request.is_disconnected():
        break
```

only checks for disconnection *after* the generator yields. If the generator is blocked on an inner `await`, `is_disconnected()` never runs.

The fix races each `generator.__anext__()` call against a disconnect watcher using `asyncio.wait(return_when=FIRST_COMPLETED)`:

- **Task A**: `generator.__anext__()` — wait for the next event
- **Task B**: poll `request.is_disconnected()` every ~500ms

If Task B wins (client disconnected), Task A is cancelled. The `CancelledError` propagates into whatever `await` the generator is suspended in, triggering cleanup via standard Python `finally` blocks — closing the inner HTTP connection.

The disconnect watcher is a single long-lived task reused across iterations, polling a boolean flag check every 500ms. `asyncio.wait` is standard event-loop machinery. Both are negligible compared to the HTTP I/O the generator performs.

None.

## How Has This Been Tested?

Added 5 unit tests that call `run_stream` directly with a `MockRequest` (controllable `is_disconnected()`), bypassing the HTTP endpoint for deterministic disconnect timing:

- **Disconnect cancels blocked generator** — the core leak scenario. Generator yields one event then blocks indefinitely on `asyncio.Event.wait()` (simulating an inner HTTP connection). After the first event, disconnect is triggered. Asserts that `CancelledError` propagated into the blocked `await` and the generator's `finally` block ran.
- **Normal exhaustion** — generator yields 3 events and finishes naturally. Regression test.
- **Disconnect between yields** — disconnect after first event when generator is not blocked. Asserts only first event received.
- **Generator exception** — generator raises `RuntimeError`. Asserts error SSE event is produced.
- **ReconnectException** — asserts reconnect SSE event is produced.

Existing 12 endpoint-level tests continue to pass, covering the HTTP integration path.

## PR Checklist:

N/A — backend-only change.

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+- Fixed StreamVariable inner HTTP connections not being cancelled when the client disconnects, preventing connection leaks when generators are blocked on upstream requests
+
 ## 1.28.3
 
 - Fixed StreamVariable SSE connections accumulating when dependency values changed, causing stale data fetching and degraded performance
--- a/packages/dara-core/dara/core/interactivity/stream_variable.py
+++ b/packages/dara-core/dara/core/interactivity/stream_variable.py
@@ -315,6 +315,12 @@ def _validate_event_mode(event: StreamEvent, key_accessor: str | None) -> None:
         )
 
 
+async def _wait_for_disconnect(request: Request, interval: float = 0.5):
+    """Poll request.is_disconnected() until the client disconnects."""
+    while not await request.is_disconnected():
+        await asyncio.sleep(interval)
+
+
 async def run_stream(
     entry: StreamVariableRegistryEntry, request: Request, values: list[Any], store: CacheStore, task_mgr: TaskManager
 ):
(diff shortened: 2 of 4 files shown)


krzysztof-causalens (approved):
nice find!

PR #646, opened 2026-06-12
Improvement: DO-11698 Expose more controls around the dara task pool

sam-causalens:
## Motivation and Context
Expose a couple of extra options from the dara worker pool so that we can keep more warm workers around if there is a startup cost to them. Also makes the timeout configureable so extra workers can be kept for more than 5 seconds.

## Implementation Description
* Exposes a min worker count as well which allows developers to keep a warm pool of workersaround if their spin up time takes a while.
* Added tests and changelog entry

No

## How Has This Been Tested?
Locally

## PR Checklist:

### DIFF
--- a/packages/dara-core/dara/core/main.py
+++ b/packages/dara-core/dara/core/main.py
@@ -177,10 +177,14 @@ def _start_application(config: Configuration):
                 # Default to number of CPUs - 1 (with minimum of 1)
                 cpu_count = get_cpu_count()
                 max_workers = int(os.environ.get('DARA_POOL_MAX_WORKERS', max(1, cpu_count - 1)))
+                worker_timeout = float(os.environ.get('DARA_POOL_WORKER_TIMEOUT', '5'))
+                min_workers = int(os.environ.get('DARA_POOL_MIN_WORKERS', '0'))
                 dev_logger.info(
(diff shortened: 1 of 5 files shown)


packages/dara-core/dara/core/main.py
```diff
@@ -177,17 +177,23 @@ async def lifespan(app: FastAPI):
                 # Default to number of CPUs - 1 (with minimum of 1)
                 cpu_count = get_cpu_count()
                 max_workers = int(os.environ.get('DARA_POOL_MAX_WORKERS', max(1, cpu_count - 1)))
+                worker_timeout = float(os.environ.get('DARA_POOL_WORKER_TIMEOUT', '5'))
+                min_workers = int(os.environ.get('DARA_POOL_MIN_WORKERS', '0'))
```
krzysztof-causalens, line 181: We should probably include those in the docs somewhere, had a quick look and even the current `MAX_WORKERS` env doesn't come up in the doc search 🙈 otherwise users (or agents) won't know those are available really.

I think we do have a section about "Heavy CPU Calculations" in the "interactivity" docs so maybe a quick mention there could work?

krzysztof-causalens (approved):
Looks good, thanks!

PR #647, opened 2026-06-12
Fix: DO-11699 avoid Link resuspension with deferred fallback

krzysztof-causalens:
## Motivation and Context

`Link` used `JSON.stringify(to)` as the key for the subtree that calls `useResolvedTo`. When `to` was a derived value, ordinary value changes could remount that subtree even when the param hook structure had not changed. That meant dependent variables such as dynamic styles could lose their deferred stale value and briefly resuspend despite `suspend_render=False`.

## Implementation Description

This changes the `Link` key to represent the hook structure of `to.params` instead of the resolved runtime value. Literal param value changes now keep the same key, while changes that affect hook order or variable identity still remount the `useResolvedTo` subtree.

A reusable `getVariableHookSignature` utility captures variable kind, uid, and nested variable-bearing fields. `Condition` is also typed with `__typename: 'Condition'` to match the runtime guard.

None.

## How Has This Been Tested?

- `pnpm --dir /Users/kbiel/code/dara/packages/dara-core exec vitest --run tests/js/variable-hook-signature.spec.tsx`
- `pnpm --dir /Users/kbiel/code/dara/packages/dara-core exec eslint js/components/link.tsx js/shared/utils/variable-hook-signature.ts js/shared/utils/index.tsx tests/js/variable-hook-signature.spec.tsx`
- `pnpm --dir /Users/kbiel/code/dara/packages/dara-core exec prettier --check js/components/link.tsx js/shared/utils/variable-hook-signature.ts js/shared/utils/index.tsx js/types/core.tsx tests/js/variable-hook-signature.spec.tsx`
- `pnpm --dir /Users/kbiel/code/dara/packages/dara-core exec tsc --noEmit --pretty false`
- `pnpm --dir /Users/kbiel/code/dara exec lerna run lint --scope @darajs/core`
- `pnpm --dir /Users/kbiel/code/dara exec lerna run format:check --scope @darajs/core`

## PR Checklist:

N/A

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,10 @@
 title: Changelog
 ---
 
+## NEXT
+
+- Fixed a case where the `Link` component would resuspend despite `suspend_render=False`.
+
 ## 1.28.4
 
 - Fixed StreamVariable inner HTTP connections not being cancelled when the client disconnects, preventing connection leaks when generators are blocked on upstream requests
--- a/packages/dara-core/js/components/link.tsx
+++ b/packages/dara-core/js/components/link.tsx
@@ -8,6 +8,7 @@ import { DisplayCtx } from '@/shared/context';
 import DynamicComponent from '@/shared/dynamic-component/dynamic-component';
 import { useVariable } from '@/shared/interactivity';
 import useComponentStyles from '@/shared/utils/use-component-styles';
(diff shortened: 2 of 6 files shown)


momustafa-causalens (approved):
thanks for looking into this and sorting it out! much appreciate 🙏

PR #649, opened 2026-06-16
Feat: DO-11744 - implicit batch updates in @action

momustafa-causalens:
## Motivation and Context

When an `@action` handler calls `ctx.update()` multiple times, each call sends a separate WebSocket message to the client. Each message triggers an independent React re-render cycle. For actions that update N variables, this means N full re-render passes before the UI settles. With large component trees this can add up to seconds of wasted client-side work, and DerivedVariables that depend on multiple updated variables resolve prematurely with partially-updated state, causing wasted server round-trips.

This PR implements implicit batching so that all actions from a single `@action` execution are collected and applied as one atomic React render on the client. The mental model is simple: everything is delivered at the end of the action, unless you call `ctx.flush()`.

## Implementation Description

**Server side:**

- Added `BatchStart` / `BatchEnd` `ActionImpl` marker subclasses that serialize as regular action messages (`__typename: 'ActionImpl'`, `name: 'BatchStart'` / `'BatchEnd'`).
- `_stream_action` now sends a `BatchStart` before running the handler and a `BatchEnd` after (just before the existing `None` sentinel). This wraps every action execution in an implicit batch with zero code changes required by app developers. A `batch` parameter (default `True`) controls whether framing markers are emitted.
- Added `ctx.flush()` on `ActionCtx` as an escape hatch: it sends `BatchEnd` + `BatchStart` to end the current batch early and start a new one, for actions that need progressive UI feedback (e.g. showing a loading spinner before slow work).
- `execute_action_sync` (used for route `on_load` actions) passes `batch=False` since its results are accumulated and sent in one HTTP response, making client-side batch framing unnecessary.

**Client side:**

* The `executeAction` function in `use-action.tsx` now recognises batch framing markers. Between `BatchStart` and `BatchEnd`, ActionImpl objects are **buffered** without executing their handlers. Nothing runs until `BatchEnd`.
* On `BatchEnd`, all buffered handlers are executed sequentially with a batching `ActionContext`. Handlers' `set` / `reset` calls are applied to a transaction snapshot (via Recoil's `snapshot.map`) rather than to the live store. Handlers read from this transaction snapshot, so reads within the batch see writes from earlier handlers -- including correct defaults after `reset` (since `MutableSnapshot.reset` knows atom defaults natively). `eventBus.publish` calls are deferred.
* After all handlers have run, final values of touched atoms are read from the transaction snapshot and committed atomically via a single `transact_UNSTABLE` call. This avoids `gotoSnapshot` which would replace the entire store state and could clobber concurrent modifications (e.g. server variable pushes, stream updates). Deferred callbacks (event bus publishes) run after the commit.
* All action types are batched uniformly -- there is no distinction between "state-mutating" and "side-effect" actions. This keeps the mental model simple ("everything is batched") and ensures custom `ActionImpl` subclasses are also covered.

**Why implicit batching is safe:** `ctx.update()` is fire-and-forget with no client acknowledgement, DerivedVariable values are resolved before the action runs (not during), and the server has no access to client-side Recoil state. No existing action can depend on intermediate client-side reactions to its own prior updates.

No

## How Has This Been Tested?

- 4 new backend integration tests covering batch framing, marker filtering, `ctx.flush()` splitting, and empty batches
- All 728 existing Python tests pass (including updated `test_websocket.py::test_action_handler_error`)
- TypeScript lint and type-check pass (`pnpm lerna run lint --scope=@darajs/core`)

## PR Checklist

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -2,6 +2,11 @@
 title: Changelog
 ---
 
+## NEXT
+
+- Added implicit batching for `@action` handlers: all `ctx.update()`, `ctx.trigger()`, and `ctx.reset_variables()` calls within a single action execution are now collected and applied as one atomic React render on the client, instead of N separate re-renders for N updates.
+- Added `ctx.flush()` escape hatch to `ActionCtx` for actions that need progressive UI feedback (e.g. showing a loading state before slow work). Calling `flush()` delivers all buffered updates immediately and starts a new batch.
+
 ## 1.28.5
 
 - Added `DARA_POOL_MIN_WORKERS` env var to keep a minimum number of warm task pool workers alive at all times, avoiding repeated cold worker spawns for apps with expensive task-module imports.
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -4,7 +4,7 @@ title: Changelog
 
 ## NEXT
 
-- Added implicit batching for `@action` handlers: all `ctx.update()`, `ctx.trigger()`, and `ctx.reset_variables()` calls within a single action execution are now collected and applied as one atomic React render on the client, instead of N separate re-renders for N updates.
+- Added implicit batching for `@action` handlers: all actions within a single action execution are now collected and applied as one atomic React render on the client, instead of N separate re-renders for N updates.
 - Added `ctx.flush()` escape hatch to `ActionCtx` for actions that need progressive UI feedback (e.g. showing a loading state before slow work). Calling `flush()` delivers all buffered updates immediately and starts a new batch.
 
 ## 1.28.5
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -4,7 +4,7 @@ title: Changelog
 
 ## NEXT
 
-- Added implicit batching for `@action` handlers: all actions within a single action execution are now collected and applied as one atomic React render on the client, instead of N separate re-renders for N updates.
+- Added implicit batching for `@action` handlers: all actions within a single action execution are now collected and applied atomically, so dependent `DerivedVariable` chains only recompute once against the final consistent state.
 - Added `ctx.flush()` escape hatch to `ActionCtx` for actions that need progressive UI feedback (e.g. showing a loading state before slow work). Calling `flush()` delivers all buffered updates immediately and starts a new batch.
 
 ## 1.28.5
--- a/packages/dara-core/dara/core/internal/execute_action.py
+++ b/packages/dara-core/dara/core/internal/execute_action.py
@@ -31,6 +31,8 @@ from dara.core.interactivity.actions import (
     BOUND_PREFIX,
     ActionCtx,
     ActionImpl,
+    BatchEnd,
+    BatchStart,
 )
 from dara.core.internal.cache_store import CacheStore
 from dara.core.internal.dependency_resolution import resolve_dependency
@@ -93,16 +95,28 @@ async def _stream_action(
     - The handler itself
     - A stream consumer which sends the results to the frontend
 
+    All updates from the action handler are implicitly batched: a BatchStart marker is sent
+    before the handler runs, and a BatchEnd marker is sent after it completes. The client
+    buffers state-mutating actions (UpdateVariable, ResetVariables, TriggerVariable) received
+    between these markers and applies them atomically in a single React render cycle.
+
+    The handler can call ``ctx.flush()`` to end the current batch early and start a new one,
+    for progressive UI feedback (e.g. showing a loading state before slow work).
+
     :param handler: the action handler to execute
     :param ctx: the action context to use
     :param values: the resolved values to pass to the handler
     """
     try:
+        # Open an implicit batch before the handler runs
+        await ctx._on_action(BatchStart())
         async with anyio.create_task_group() as tg:
             # Execute the handler and a stream consumer in parallel
             tg.start_soon(partial(_execute_action, _on_error=_on_error), handler, ctx, values)
             tg.start_soon(ctx._handle_results)
     finally:
+        # Close the implicit batch, then send the sentinel to signal completion
+        await ctx._on_action(BatchEnd())
         # None is treated as a sentinel value to stop waiting for new actions to come in on the client
         await ctx._on_action(None)
 
@@ -131,9 +145,11 @@ async def execute_action_sync(
 
     results = []
 
-    # Construct a context which handles action messages by accumulating them in an array
+    # Construct a context which handles action messages by accumulating them in an array.
+    # Batch framing markers are filtered out since on_load results are sent to the client
+    # all at once (batching is handled at the WS streaming level, not here).
     async def handle_action(act_impl: ActionImpl | None):
-        if act_impl is not None:
+        if act_impl is not None and not isinstance(act_impl, (BatchStart, BatchEnd)):
             results.append(act_impl)
 
     ctx = ActionCtx(inp, handle_action)
--- a/packages/dara-core/dara/core/internal/execute_action.py
+++ b/packages/dara-core/dara/core/internal/execute_action.py
@@ -103,7 +103,8 @@ async def _stream_action(
     When ``batch`` is True (the default), all actions from the handler are implicitly
     batched: a BatchStart marker is sent before the handler runs, and a BatchEnd marker
     is sent after it completes. The client buffers all actions received between these
-    markers and applies them atomically in a single React render cycle.
+    markers and applies them atomically, so dependent variable chains only recompute once
+    against the final consistent state.
 
     The handler can call ``ctx.flush()`` to end the current batch early and start a new one,
     for progressive UI feedback (e.g. showing a loading state before slow work).
--- a/packages/dara-core/dara/core/internal/execute_action.py
+++ b/packages/dara-core/dara/core/internal/execute_action.py
@@ -87,7 +87,12 @@ async def _execute_action(
 
 
 async def _stream_action(
-    handler: Callable, ctx: ActionCtx, _on_error: Literal['raise', 'notify'] = 'notify', **values: Mapping[str, Any]
+    handler: Callable,
+    ctx: ActionCtx,
+    _on_error: Literal['raise', 'notify'] = 'notify',
+    *,
+    batch: bool = True,
+    **values: Mapping[str, Any],
 ):
     """
     Run the action handler and stream the results to the frontend.
@@ -95,28 +100,33 @@ async def _stream_action(
     - The handler itself
     - A stream consumer which sends the results to the frontend
 
-    All updates from the action handler are implicitly batched: a BatchStart marker is sent
-    before the handler runs, and a BatchEnd marker is sent after it completes. The client
-    buffers state-mutating actions (UpdateVariable, ResetVariables, TriggerVariable) received
-    between these markers and applies them atomically in a single React render cycle.
+    When ``batch`` is True (the default), all actions from the handler are implicitly
+    batched: a BatchStart marker is sent before the handler runs, and a BatchEnd marker
+    is sent after it completes. The client buffers all actions received between these
+    markers and applies them atomically in a single React render cycle.
 
     The handler can call ``ctx.flush()`` to end the current batch early and start a new one,
     for progressive UI feedback (e.g. showing a loading state before slow work).
 
+    When ``batch`` is False, no batch framing markers are emitted. This is used by
+    ``execute_action_sync`` where results are accumulated and sent in one HTTP response,
+    making client-side batching unnecessary.
+
     :param handler: the action handler to execute
     :param ctx: the action context to use
+    :param batch: whether to emit batch framing markers (default True)
     :param values: the resolved values to pass to the handler
     """
     try:
-        # Open an implicit batch before the handler runs
-        await ctx._on_action(BatchStart())
+        if batch:
+            await ctx._on_action(BatchStart())
         async with anyio.create_task_group() as tg:
             # Execute the handler and a stream consumer in parallel
             tg.start_soon(partial(_execute_action, _on_error=_on_error), handler, ctx, values)
             tg.start_soon(ctx._handle_results)
     finally:
-        # Close the implicit batch, then send the sentinel to signal completion
-        await ctx._on_action(BatchEnd())
+        if batch:
+            await ctx._on_action(BatchEnd())
         # None is treated as a sentinel value to stop waiting for new actions to come in on the client
         await ctx._on_action(None)
 
@@ -145,11 +155,9 @@ async def execute_action_sync(
 
     results = []
 
-    # Construct a context which handles action messages by accumulating them in an array.
-    # Batch framing markers are filtered out since on_load results are sent to the client
-    # all at once (batching is handled at the WS streaming level, not here).
+    # Construct a context which handles action messages by accumulating them in an array
     async def handle_action(act_impl: ActionImpl | None):
-        if act_impl is not None and not isinstance(act_impl, (BatchStart, BatchEnd)):
+        if act_impl is not None:
             results.append(act_impl)
 
     ctx = ActionCtx(inp, handle_action)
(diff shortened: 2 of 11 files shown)


packages/dara-core/docs/getting-started/actions.mdx
```diff
@@ -53,6 +53,19 @@ The example above shows how to use the `@action` decorator to create an action t
...
 Similarly to `py_component`s, the `@action`-decorated function can take a mixture of regular Python variables and Dara `Variable`-based arguments in any combination. The `@action` decorator will automatically resolve the `Variable`-based arguments so your function will receive the current value of the `Variable` instead of the `Variable` instance.
 
+:::tip Automatic Batching
+
+All `ctx.update()`, `ctx.trigger()`, and `ctx.reset_variables()` calls within a single `@action` execution are **automatically batched**.
+Instead of each call triggering a separate React re-render on the client, the framework collects all state-mutating updates and applies them
+as a single atomic operation when the action finishes. This means an action that updates N variables only causes **one** React render cycle
+instead of N, which can significantly improve performance for actions that update multiple variables.
+
+Non-state-mutating methods like `ctx.notify()`, `ctx.navigate()`, `ctx.download_file()` etc. are not affected by batching and are applied immediately.
```
krzysztof-causalens: What's the reasoning for this? It complicates the mental model vs just "everything is applied at the end unless you flush" which could be simpler. You could argue `notify` is usually immediate feedback but so could be `update()` so we can't really make the choice for users?

packages/dara-core/dara/core/internal/execute_action.py
```diff
@@ -131,9 +145,11 @@ async def execute_action_sync(
 
     results = []
 
-    # Construct a context which handles action messages by accumulating them in an array
+    # Construct a context which handles action messages by accumulating them in an array.
+    # Batch framing markers are filtered out since on_load results are sent to the client
+    # all at once (batching is handled at the WS streaming level, not here).
     async def handle_action(act_impl: ActionImpl | None):
-        if act_impl is not None:
+        if act_impl is not None and not isinstance(act_impl, (BatchStart, BatchEnd)):
```
krzysztof-causalens: I don't get why filter them out here? I get that we buffer everything anyway but there should be no harm in sending batches and it just complicates the code here?

packages/dara-core/js/shared/interactivity/use-action.tsx
```diff
@@ -29,6 +29,111 @@ import { normalizeRequest } from '../utils/normalization';
 import { cleanKwargs, getOrRegisterPlainVariable, resolveVariable } from './internal';
 import { useVariable } from './use-variable';
 
+// Names of batch framing markers sent by the server
+const BATCH_START = 'BatchStart';
+const BATCH_END = 'BatchEnd';
+
+// Action names whose handlers mutate Recoil state (ctx.set / ctx.reset) and should
+// be buffered during a batch so they can be applied atomically in a single render cycle.
+const STATE_MUTATING_ACTIONS = new Set(['UpdateVariable', 'ResetVariables', 'TriggerVariable']);
```
krzysztof-causalens: Again alternatively, we just buffer all action executions and this becomes simpler. Otherwise custom actions aren't buffered as well and there might be cases where you want that

krzysztof-causalens (changes requested):
Nice, I was thinking about doing something similar for a while.

I'm somewhat worried about implicit batching as it changes the behaviour slightly. It's probably the right default but it could possibly break long-running actions, e.g. imagine existing actions that do:

```python
await ctx.update(...)
await ctx.run_task(...)
await ctx.update(...)
```

Now the initial update won't show unless developer adds a `flush()`. I do think it's the better behaviour but it's technically breaking :/ 

We might be okay with that if we make this a minor release, announce the change. Technically it was an implied behaviour so you could argue relying on previous one was not guaranteed etc.

Anyway left some questions below, I think we could simplify if we just batch all action executions.

packages/dara-core/docs/getting-started/actions.mdx
momustafa-causalens: i think it was motivated by my initial tunnel vision towards fixing the react re-renders, notify(), navigate(), etc. don't call ctx.set() on Recoil atoms, so they don't trigger React re-renders of the component tree. Batching them would delay things like notifications and navigations for no benefit since they don't contribute to the render-batching problem. but i agree we cant make the choice for the user cz they might use update as a way to provide immediate feedback too

krzysztof-causalens: There is no benefit but it simplifies the code and mental model, plus any custom actions immediately work with the system. There isn't much downside, for most actions that are reasonably quick they would just be delayed slightly until end of action; if you have something in the middle that would take some time you probably want to `flush` before that anyway, and with current API you also have to think whether you did `update` or `notify` and swapping from one to the other changes behaviour.
I just think the model is more consistent if we batch everything.

packages/dara-core/dara/core/internal/execute_action.py
momustafa-causalens: cz otherwise resolveActionImpl would fail to find a handler for them (ACTION_HANDLER_BY_NAME['BatchStart'] doesn't exist) and throw an UnhandledActionError.

it either doing it here or in _stream_action we skip batch framing when called from execute_action_sync, since batching is irrelevant there but that seemed like a more convoluted change?

ill have a think and see if theres a more elegant solution

packages/dara-core/docs/getting-started/actions.mdx
krzysztof-causalens: The other big alternative is exposing the batch API directly, e.g. `await ctx.batch_start()` and `await ctx.batch_end()` and not batch by default (but that's if we want to preserve current behaviour)

packages/dara-core/dara/core/internal/execute_action.py
krzysztof-causalens: I think we can just define a client-side handler for this that's a noop instead?

krzysztof-causalens: Idk not too elegant either as it shouldn't reach that point in the stack and it should be handled higher up. Thinking about it maybe it should be an equivalent of a WS "control frame", so treated as a special kind of an action impl that's handled on a protocol level. Would require some flag on the impl object I guess so not perfectly clean either

packages/dara-core/docs/getting-started/actions.mdx
```diff
@@ -53,6 +53,17 @@ The example above shows how to use the `@action` decorator to create an action t
 
 Similarly to `py_component`s, the `@action`-decorated function can take a mixture of regular Python variables and Dara `Variable`-based arguments in any combination. The `@action` decorator will automatically resolve the `Variable`-based arguments so your function will receive the current value of the `Variable` instead of the `Variable` instance.
 
+:::tip Automatic Batching
+
+All actions within a single `@action` execution are **automatically batched**.
+Instead of each action being applied to the client independently, the framework collects them and applies them
+as a single atomic operation when the action finishes. This means an action that updates N variables only causes **one** React render cycle
+instead of N, which can significantly improve performance for actions that update multiple variables.
```
krzysztof-causalens: I'm not sure I would mention `React` here, we're trying not to assume reader has any frontend experience. I would explain it in Dara terms instead - e.g. it will only cause related variable chains to recompute once etc.

packages/dara-core/dara/core/internal/execute_action.py
```diff
@@ -85,24 +87,46 @@ async def _execute_action(
...
     """
     Run the action handler and stream the results to the frontend.
     Executes two tasks in parallel:
     - The handler itself
     - A stream consumer which sends the results to the frontend
 
+    When ``batch`` is True (the default), all actions from the handler are implicitly
+    batched: a BatchStart marker is sent before the handler runs, and a BatchEnd marker
+    is sent after it completes. The client buffers all actions received between these
+    markers and applies them atomically in a single React render cycle.
```
krzysztof-causalens: Same here, wouldn't mention React

packages/dara-core/docs/getting-started/actions.mdx
```diff
@@ -489,6 +503,52 @@ def task_page():
 config.router.add_page(path='task', content=task_page)
 ```
 
+### `flush`
+
+```python
+async def flush()
+```
+
+`flush` ends the current implicit batch and starts a new one. All actions buffered since the start of the action (or the last `flush`) are delivered to the client immediately and applied as a single atomic React render.
```
krzysztof-causalens: Same here

packages/dara-core/docs/getting-started/actions.mdx
```diff
@@ -489,6 +503,52 @@ def task_page():
...
+result = Variable(None)
+
+@action
+async def load_data(ctx: action.Ctx):
+    # Show loading state immediately
+    await ctx.update(loading, True)
+    await ctx.flush()
+
+    # ... slow work (e.g. API call, computation) ...
+    data = expensive_computation()
```
krzysztof-causalens: I would show this as `await ctx.run_task` or just `await long_task()`, CPU bound/synchronous long-running operations shouldn't be encouraged directly in actions.

packages/dara-core/js/shared/interactivity/use-action.tsx
```diff
@@ -29,6 +29,106 @@ import { normalizeRequest } from '../utils/normalization';
...
+            batch.ops.push({ type: 'set', atom: atom as RecoilState<unknown>, value: valOrUpdater });
+        }
+    };
+}
+
+/**
+ * Create an `ActionContext` that records `set` / `reset` calls into `batch.ops`
+ * instead of applying them immediately. All other fields are forwarded from `realCtx`.
+ */
+function createBatchingContext(realCtx: ActionContext, batch: BatchState): ActionContext {
```
krzysztof-causalens: Does this need to be so complicated? I thought what we would do is just buffer all the `ActionImpl` functions, and execute them altogether within a single recoil transaction - wouldn't that be enough?

momustafa-causalens: the complexity comes from the fact that  transact_UNSTABLE is synchronous and some handlers are async. We can't just stuff all handler executions inside a single transact_UNSTABLE for example UpdateVariable with toggle uses `await ctx.snapshot.getLoadable(varAtom).toPromise()`

that being said, i think theres probably ways to simplify this still

krzysztof-causalens: Ugh annoying, yeah I can see recoil is undercooked in this regard. I guess we're basically manually transacting everything, but need to be careful to preserve the "everything is batched" behaviour. So maybe we do both - accumulate all action impls, provide a batching context, then when BatchEnd is encountered, run all async action impls and flush the batched recoil ops in order?

krzysztof-causalens: There's still an issue there if some action implementation calls `set()` then needs the result of that, it won't see it so that could break the action :(

packages/dara-core/js/shared/interactivity/use-action.tsx
```diff
@@ -33,12 +33,37 @@ import { useVariable } from './use-variable';
...
 
+/*
+ * ── Batch buffering ──────────────────────────────────────────────────────────
+ *
+ * `transact_UNSTABLE` requires a **synchronous** callback, and some action
+ * handlers are async so we cannot buffer all ActionImpl objects and execute
+ * them inside a single `transact_UNSTABLE` call.
+ *
+ * Instead we:
+ *   1. Run handlers normally as they arrive, but with a recording `ActionContext`
```
krzysztof-causalens: That would mean things like `.notify` execute immediately? I thought we established those would be buffered as well. The only way to do that is to buffer the `ActionImpl` objects as well and defer their execution until BatchEnd

packages/dara-core/js/shared/interactivity/use-action.tsx
momustafa-causalens: Good shout, I've made an update that tracks which atoms have pending writes in the current batch as dirtyAtoms and if a handler reads an atom via snapshot.getLoadable that has a pending batched write, it emits a console.warn explaining the stale read and suggesting ctx.flush() as the fix and this should only happen for certain custom ActionImpl js handlers

Aside from that, the implementation now fully buffers ActionImpl objects during a batch without executing their handlers at all. On BatchEnd, all buffered handlers are executed sequentially (supporting async) with a recording ActionContext that captures set/reset/eventBus.publish calls, then all recorded Recoil ops are committed atomically via a single transact_UNSTABLE, then deferred callbacks run.

packages/dara-core/js/shared/interactivity/use-action.tsx
```diff
@@ -93,21 +115,48 @@ interface BatchState {
...
+            if (prop === 'getLoadable' || prop === 'getPromise') {
+                const original = Reflect.get(target, prop, receiver) as (...args: unknown[]) => unknown;
+                return function warnOnStaleRead(recoilValue: RecoilState<unknown>, ...rest: unknown[]) {
+                    if (batch.dirtyAtoms.has(recoilValue)) {
+                        // eslint-disable-next-line no-console
+                        console.warn(
+                            `[Dara] Reading atom "${String(recoilValue.key)}" via snapshot inside a batch, ` +
+                                'but this atom has a pending batched write. The snapshot reflects pre-batch state ' +
+                                'so the value returned will be stale. Use ctx.flush() before this read if you need ' +
+                                'the updated value.'
```
krzysztof-causalens: If the issue is in a custom action, the developer of it can't really `flush` inside of it so the suggestion here is wrong. I wonder if we could somehow keep track of the updated values of the atom so we can return the cached new value instead?... 

Kind of like within db transactions you can see updated values, if that makes sense. I think we have everything we need to do that? We would just need to hold a map of atom->updated value, have a look if that's feasible.
It would be a little complicated as we would need to defer the actual recoil application while keeping the "in-transaction" state updated "live" if that makes sense but I think it's doable

krzysztof-causalens: So effectively instead of ` batch.dirtyAtoms.add(atom);` you do something like `batch.atomsCache[key] = value` (or `valueUpdater(previousValue)`), 
then `getLoadable/Promise` becomes:

```ts
if (key in batch.atomsCache) {
   return batch.atomsCache[key];
}

return original();
```
conceptually at least

momustafa-causalens: Okay so then we'd add a localState: Map<atom, value> that tracks the latest value written to each atom within the batch. batchSet stores the resolved value in the map (for updaters, it applies the function against the current local value or snapshot), and the wrapped snapshot.getLoadable/getPromise checks the map first before falling through to the real snapshot?

One edge case though: reset. In Recoil, reset means "restore to the atom's default value", but we don't have access to the atom's default from here. The options I see:

- Delete from localState on reset, let reads fall through to the real snapshot. This means a read after a reset within the same batch would return the pre-batch value (which might differ from the atom's default if it was set before the action started).

- Store a sentinel in localState for reset, and have getLoadable fall through to the real snapshot when it sees the sentinel.

what do you think? how likely is this?

krzysztof-causalens: I guess if you do `ctx.reset()` into `ctx.toggle()`, that would potentially read the wrong value ugh, not very likely but easy to reproduce.

Not sure what the most expected behaviour would be then. You would kind of want it to function somehow.
You could in theory read the default off of the `variable.default` field - the only edge case then becomes if the variable itself has a DerivedVariable as a default (which is allowed), then you would need to read the current DV value. Maybe it's possible?
We do have a reverse lookup possibility from atom->its variable definition, so maybe the above is fine?

krzysztof-causalens (commented):
Nice! I think this looks much better.
Could we add some client-side coverage for the useAction behaviours? Simulate the happy path and couple of edge cases we can think of to ensure the snapshots etc behave as expected?

packages/dara-core/changelog.md
```diff
@@ -2,6 +2,11 @@
 title: Changelog
 ---
 
+## NEXT
+
+- Added implicit batching for `@action` handlers: all actions within a single action execution are now collected and applied as one atomic React render on the client, instead of N separate re-renders for N updates.
```
krzysztof-causalens: Let's remove the React wording here

krzysztof-causalens (approved):
LGTM, let's just update the changelog and good to merge

PR #655, opened 2026-06-26
Feat: DO-11819 - add HTTP request metrics middleware

momustafa-causalens:
## Motivation and Context

Dara already exposes Prometheus metrics on port 10000 (task runtimes, DV runtimes, cache size), but has no HTTP request-level metrics. Adding `http_requests_total` and `http_request_duration_seconds` enables the observability service to surface request rate, error rate, and latency for deployed Dara apps.

## Implementation Description

- **New `metrics/http.py`** — defines `http_requests_total` (Counter) and `http_request_duration_seconds` (Histogram) with labels `method`, `status`, `path`. A `PrometheusMiddleware` (Starlette `BaseHTTPMiddleware`) wraps every request and records both metrics. Uses the matched route pattern (e.g. `/api/components/{uid}`) for the `path` label to avoid unbounded cardinality.

- **Middleware registration in `main.py`** — gated by the same conditions as the existing metrics server (`DARA_DISABLE_METRICS != TRUE` and `DARA_TEST_FLAG` not set), so tests are unaffected.

- **Exports added to `metrics/__init__.py`** — `HTTP_REQUESTS_TOTAL` and `HTTP_REQUEST_DURATION_SECONDS` are available for downstream code if needed.

No changes to the existing `/status` endpoint, metrics server startup, or `prometheus_client` dependency — all already in place.

> [!NOTE]
> Not sure if we need a changelog update for this sort of change

None.

## How Has This Been Tested?

- Targeted test suite passes (70 passed). The middleware is not registered during tests (`DARA_TEST_FLAG` guard), so existing tests are unaffected.
- Import verification: `from dara.core.metrics.http import PrometheusMiddleware` succeeds.

## PR Checklist:

N/A — backend-only change. Metrics are visible via `curl http://localhost:10000/metrics`.

### DIFF
--- a/packages/dara-core/dara/core/metrics/http.py
+++ b/packages/dara-core/dara/core/metrics/http.py
@@ -18,7 +18,7 @@ limitations under the License.
 import time
 
 from prometheus_client import Counter, Histogram
-from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
 from starlette.requests import Request
 from starlette.responses import Response
 
@@ -43,7 +43,7 @@ class PrometheusMiddleware(BaseHTTPMiddleware):
     label to avoid unbounded cardinality from dynamic path segments.
     """
 
-    async def dispatch(self, request: Request, call_next) -> Response:  # type: ignore[override]
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
         start = time.monotonic()
         response = await call_next(request)
         duration = time.monotonic() - start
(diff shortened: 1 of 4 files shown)


packages/dara-core/dara/core/metrics/http.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class PrometheusMiddleware(BaseHTTPMiddleware):
+    """Middleware that records HTTP request count and duration as Prometheus metrics.
+
+    Uses the matched Starlette route pattern (e.g. ``/api/components/{uid}``) for the ``path``
+    label to avoid unbounded cardinality from dynamic path segments.
+    """
+
+    async def dispatch(self, request: Request, call_next) -> Response:  # type: ignore[override]
```
krzysztof-causalens: can we just type this properly?

krzysztof-causalens (commented):
Could we add a brief changelog entry, just say that we extended metrics to include http stats

krzysztof-causalens (approved):

PR #656, opened 2026-06-29
Feat: DO-11833 pluggable auth session backend storage, preserve auth in local dev with `--reload`

krzysztof-causalens:
## Goal

Keep authenticated local development sessions alive across backend reloads while preserving existing production restart behavior by default.

## Motivation and Context

Dara recently moved browser auth sessions to server-side storage. That keeps raw auth token material out of browser cookies, but the previous in-memory-only store meant every `dara start --reload` restart logged users out. This is especially painful during local development and agent-driven reload loops.

This PR adds a local file-backed auth session option and changes the automatic local reload default so reloads preserve browser sessions. It also moves local development signing-key persistence out of generated project `.env` files and into Dara-owned user cache storage.

## Non-Goals

This PR does not add a database-backed or shared-cache auth session backend. It also does not make file-backed sessions a distributed production recommendation; the file backend is local/single-host storage.

## Constraints And Invariants

- Existing non-reload runtime behavior remains in-memory by default, so existing deployments keep losing sessions on restart unless they explicitly configure another backend.
- Docker and production modes do not implicitly switch to file-backed auth sessions.
- File-backed sessions store raw auth token material and must be treated as sensitive local server-side storage.
- Missing production `JWT_SECRET` remains a compatibility fallback, but now logs a loud warning.

### Approach

Auth session storage is now pluggable through `config.auth_session_backend`. The default value is an auto factory that resolves once during application startup: local reload uses `FileAuthSessionBackend`, while normal/docker/production modes use `InMemoryAuthSessionBackend`.

The file backend stores one complete session snapshot per file, hashes opaque browser session tokens for filenames, writes via a temp file plus `os.replace`, rejects malformed/oversized/unreadable files as absent, and removes session files on revoke/logout.

Local development `JWT_SECRET` handling now resolves through a dedicated signing-key helper. Explicit settings/env/`.env` values still win. If no secret is configured in local development, Dara reuses a generated key in the platform user cache. Production/docker missing-secret behavior remains compatible but warns.

### Walkthrough

- `packages/dara-core/dara/core/auth/session_store.py` contains the backend protocol, memory backend, file backend, auto resolver, and startup registry helpers.
- `packages/dara-core/dara/core/main.py` installs the resolved backend in lifespan startup so auth routes use the configured backend for the process lifetime.
- `packages/dara-core/dara/core/configuration.py` exposes `config.auth_session_backend` for explicit backend objects or factories.
- `packages/dara-core/dara/core/internal/signing_key.py` handles local dev signing-key reuse and production/docker warnings.
- `packages/dara-core/tests/python/test_auth_reload_e2e.py` black-box tests restart behavior through real uvicorn subprocesses.
- `packages/dara-core/docs/advanced/authentication.mdx` documents the user-facing storage and signing-key behavior.

<details>
<summary>Additional Design Context</summary>

### Django Precedent

This follows the same broad shape as Django's session backend model: Django supports multiple session engines, including a file-based session engine that stores one session per file under a configurable directory. Dara borrows that backend shape for local durability while keeping the default runtime behavior conservative.

The security model is stricter for Dara's file backend. Django's built-in auth stores user/session metadata by default, while Dara's auth session files may contain raw Dara-issued auth tokens and refresh tokens. That is why the backend hashes filenames, stores files outside the repository by default, uses owner-only permissions where supported, and documents file storage as sensitive local server-side storage.

### Design Choices

- `--reload` changes the automatic default to file storage because process memory cannot survive Python reloads, but explicit `config.auth_session_backend = InMemoryAuthSessionBackend()` still wins.
- Production/docker modes keep memory as the automatic default so existing deployments do not silently start persisting auth material to disk.
- The file backend avoids a central index file and stores one complete snapshot per session file, so reads and writes can be simple and local-filesystem friendly.
- `set` updates only an existing session and returns `False` for missing sessions, so refresh cannot recreate a session that was logged out or revoked.
- Local development signing keys live in the user cache instead of generated project `.env` files, avoiding dirty worktrees while preserving reload continuity.

### Deferred Work

A database, Redis, Valkey, or other shared-cache auth session backend would be the right path for production deployments that need durable/shared sessions. This PR leaves that as a future backend implementation rather than making the local file backend the distributed production answer.

</details>

- Added `platformdirs` so Dara can store local development signing keys in the platform user cache directory instead of mutating project files.

## How Has This Been Tested?

- `cd packages/dara-core && poetry run ruff format --check tests/python/test_auth_reload_e2e.py`
- `cd packages/dara-core && poetry run ruff check tests/python/test_auth_reload_e2e.py`
- `cd packages/dara-core && DARA_TEST_FLAG=True poetry run pytest tests/python/test_auth_reload_e2e.py -vv`
- `cd packages/dara-core && DARA_TEST_FLAG=True poetry run pytest tests/python/test_auth_reload_e2e.py tests/python/test_settings.py tests/python/test_auth.py tests/python/test_oidc_auth.py tests/python/test_auth_session_store.py`
- `poetry anthology run lint`

Verified locally in `demo_app` - using with `--reload` now fully preserves local sessions, doesn't cause log-outs as it did previously.

## Risks And Rollback

The main behavioral risk is that local reload now persists auth session files under the platform temp directory. The files contain raw auth token material, so the docs and runtime warning call this out explicitly. Rollback is straightforward: restore the auto backend default to memory and remove the file backend selection path.

The signing-key change affects local development fallback behavior. Explicit `JWT_SECRET` values still win, so production deployments can avoid the fallback entirely by setting the secret through their deployment environment.

## Reviewer Guidance

Please focus on backend selection semantics, file storage safety, and whether the signing-key fallback preserves compatibility without hiding production misconfiguration. The e2e subprocess test is intended to verify the behavior users actually experience across process restarts.

## PR Checklist:

N/A - backend/auth storage change only.


krzysztof-causalens:
@codex review

packages/dara-core/dara/core/auth/session_store.py
```diff
@@ -188,19 +264,483 @@ async def get(self, session_token: str) -> StoredAuthSession | None:
...
+            try:
+                entry = _build_session_entry(auth_token, token_data, refresh_token, now)
+            except ValueError:
+                self._remove_file(path)
+                raise
+
+            if not path.exists():
+                return False
+
+            self._write_entry_file(path, entry)
```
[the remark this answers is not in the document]
krzysztof-causalens: Dara is currently single-process so not an issue for now

packages/dara-core/dara/core/auth/session_store.py
```diff
@@ -188,19 +264,483 @@ async def get(self, session_token: str) -> StoredAuthSession | None:
...
+            for path in self._valid_session_files():
+                entry = self._read_entry_file(path)
+                if entry is not None and entry.retention_expires_at <= now:
+                    self._remove_file(path)
+
+
+def _is_local_reload(config: 'Configuration') -> bool:
+    is_reload = (
+        config.live_reload or os.environ.get('DARA_LIVE_RELOAD') == 'TRUE' or os.environ.get('DARA_HMR_MODE') == 'TRUE'
+    )
```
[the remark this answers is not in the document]
krzysztof-causalens: Fixed, now only --reload changes the default backend to file since with --enable-hmr backend doesn't restart.

momustafa-causalens (approved):
thanks so much for this ❤️

packages/dara-core/dara/core/auth/session_store.py
```diff
@@ -188,19 +265,478 @@ async def get(self, session_token: str) -> StoredAuthSession | None:
...
+    async def get(self, session_token: str) -> StoredAuthSession | None:
+        """
+        Return the server-side auth session for an opaque session token.
+
+        :param session_token: Opaque browser session handle.
+        """
+
+        now = time.time()
+        async with self._lock:
+            return self._get_unlocked(session_token, now)
```
sam-causalens: If I'm reading this correctly this is now re-reading this file on every request as I would imagine that verifying the token requires reading the session? That's quite a bit of overhead and would likely make this unusable on a platform deployment due to it being an NFS underneath and this being a sync file read.

Could we look into making the file reading async so it doesn't block the loop when it's being used and can we also put in an LRU cache for sessions so active sessions aren't re-read from disk all the time.

Treating the filestore as the eventual source of truth, but keeping the active sessions updated in memory would then hopefully allow us to confidently use this in production rather than this being an advised dev only option.

sam-causalens (changes requested):
Thanks Krys, looks good overall. Left one thing to think about before this goes in.

packages/dara-core/dara/core/auth/session_store.py
krzysztof-causalens: So I agree with the performance concern. I will add those two improvements (async + caching) to improve perf.

But I don't think those changes alone make the file backend the right "proper production" approach. They improve the single-process/single-host case, but they don’t solve the broader storage-contract issues:
- with multiple workers, per-process caches can become stale unless there is a shared invalidation mechanism
- shared filesystems/NFS have metadata caching, latency, and failure-mode quirks that are not a great auth-session coordination primitive
- cleanup/retention is easier and more observable in a DB or shared cache with expiry semantics, e.g. Redis TTL
- Redis/Valkey would generally be much better for hot session reads, and a DB-backed store would be a better durable source of truth

The framing I was following is similar to Django's approach (https://docs.djangoproject.com/en/6.0/topics/http/sessions/#module-django.contrib.sessions). They default to DB-backed sessions, which Dara can't do because it doesn't have a native DB integration, and they point users toward cache-backed sessions for performance, with `cached_db` when persistence matters.

So I wouldn't propose changing the docs to recommend the file backend as the production scaling story. Async + cache should make it good enough for local/demo/POC usage, including current single-worker platform deployments where session retention is desired.

I would prefer to keep the docs saying that the default memory backend is still fastest if session loss on restart is acceptable. If durable/shared sessions matter, the right long-term answer is a DB/Redis/Valkey backend.

Let me know what you think?

sam-causalens: I agree that for a multi worker running mode that it doesn't make sense, but we can't really deploy Dara in that mode at the moment so not sure the concern is as prominent as in something like Django? Definitely not perfect, but for how we use it, then it's likely better than just the in memory one?

krzysztof-causalens: That's true, with LRU cache etc the perf downside of using file backend goes down a bit. Main cost is on cache miss and the create/update/logout flows which have to hit the real FS and hold the lock on the storage. 

Even if the cost is acceptable though, not sure this should be the default though. Other aspects remain, such as that production deployments would then start writing raw auth session material to disk:
- currently this defaults to `tempdir` which is usually ephemeral though so restart persistence may not actually hold unless the path is configured
- I don't think there is a default path we could set anyway that's reasonable as it needs to be mounted 
- cleanup might be a problem? ideally you would run some process there, the backend implements `clear_expired()` but it's not wired to any bg maintenance process or similar

I would be hesitant to make this our default experience, but I can accept a middleground where we change our stance in docs to:
> file is reasonable where restart continuity is worth local-disk persistence but requires setting a mounted folder for it

so it stays opt-in outside of local `--reload` mode.

sam-causalens: Okay, happy to leave it 👍

krzysztof-causalens: Made all IO operations async, added LRU cache, and rephrased recommended usage to clearly spell out the tradeoff. File-backend is now a recommended solution if simple storage is required and the small perf penalty and disk storage is acceptable. 
Let me know if this looks okay.

sam-causalens (approved):
LGTM, thanks 🙏

PR #671, opened 2026-07-27
Improvement: DO-12330 speed up E2E CI

krzysztof-causalens:
## Motivation and Context

The E2E job is currently the slowest CI job after the Oxlint/Oxfmt tooling upgrade. It also sits behind a shared setup job, so every real test job waits for setup even when its caches are already warm.

Measured GitHub Actions results from [master run 30278046973](https://github.com/causalens/dara/actions/runs/30278046973) and [this PR's run 30283988882](https://github.com/causalens/dara/actions/runs/30283988882):

| Measurement | `master` | This PR |
| --- | ---: | ---: |
| Workflow wall time | 7m 38s | 4m 36s |
| Shared setup job | 1m 42s | removed |
| E2E job | 5m 42s | 4m 27s |
| Cypress test step | 3m 13s | 2m 40s |

The E2E job is 1m 15s faster (22%), while the workflow critical path is 3m 02s faster (40%). GitHub-hosted runner timings naturally include some run-to-run noise.

## Implementation Description

- Upgrade Cypress from 10.11.0 to 15.19.0 and fix the stricter subject handling introduced by newer Cypress releases.
- Cache the default authentication session across specs with `cy.session`, while continuing to visit each test route independently.
- Remove the shared setup job so lint, unit-test, and E2E jobs can begin immediately.
- Preserve cold-cache correctness by installing the Python environment and Cypress binary in the consuming job on cache misses.
- Let `actions/setup-node` manage the pnpm store cache and remove the redundant apt update/wget installation.
- Rely on the GitHub-hosted Ubuntu image's Cypress runtime libraries instead of downloading, installing, and caching an outdated APT dependency set on every new cache key.
- Use shallow checkouts in test jobs and let `cypress run` validate the restored binary instead of launching Cypress once in a separate verification step.

Controlled local measurements, included only to separate the Cypress and authentication effects:

| Configuration | Cypress-reported time | Local wall time |
| --- | ---: | ---: |
| Cypress 10 + cached authentication | 83s | 100.7s |
| Cypress 15 + cached authentication | 75s | 98.2s |
| Cypress 15 + original authentication flow | 84s | 104.1s |

No new dependency category. The existing Cypress development dependency is upgraded from 10.11.0 to 15.19.0.

## How Has This Been Tested?

- Ran the full 55-test E2E suite successfully on both Cypress 10.11.0 and 15.19.0.
- Re-ran the navigation compatibility spec on the final Cypress 15 state.
- Checked the affected navigation manually in a local browser.
- Ran `make lint-js`.
- Ran `pnpm lerna run format:check`.
- Verified the frozen pnpm lockfile, Cypress binary, workflow YAML, and `git diff --check`.
- Passed the complete GitHub Actions workflow without the explicit Cypress APT dependency step.
- Passed [follow-up run 30285912316](https://github.com/causalens/dara/actions/runs/30285912316) using depth-1 checkouts and direct Cypress launch from the restored binary cache.

## PR Checklist:

- [x] I have updated the appropriate changelog with a line for my changes. (Not applicable: internal CI/test tooling only.)

Not applicable; this changes test and CI infrastructure only.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,8 +15,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          fetch-depth: 0
       - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: "3.10"
@@ -48,8 +46,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          fetch-depth: 0
       - name: Install pnpm
         uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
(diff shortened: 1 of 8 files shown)


sam-causalens (approved):
LGTM

PR #672, opened 2026-07-29
Improvement: DO-12359 Detect mismatched development servers

krzysztof-causalens:
## Motivation and Context

Running `dara start --enable-hmr` against a Vite server started from another Dara project currently fails later with opaque component resolution errors. The tooling also assumed a `dist` static directory and a fixed development port, even though both need to be configurable.

This set of improvements:
- makes it easier to work on multiple dara apps (since they can run on different ports now)
- makes it harder to mess up and easier to spot the issue for both humans and agents

## Implementation Description

- Load the application configuration in `dara dev` and use its configured static files directory.
- Add a development-only identity handshake between the backend and Vite. A mismatched, unavailable, malformed, or older server renders a single Dara-styled diagnostic page that identifies the expected project and retries automatically.
- Add `dara start --dev-port` and `dara dev --port`, share the selected origin with Vite, and enable `strictPort` so the two processes cannot silently disagree.
- Skip the handshake entirely in production and UMD/AutoJS modes.

The identity payload is parsed at the HTTP boundary and transported as base64url-encoded JSON so the diagnostic can show useful project paths rather than only a hash.

None.

## How Has This Been Tested?

- 33 focused Dara Core Python tests covering CLI configuration, identity parsing/comparison, and application rendering.
- 4 focused Vitest tests covering the Vite development server plugin.
- `poetry anthology run lint`
- `make lint-js`
- `pnpm lerna run format:check`
- Manual smoke test in `packages/demo-app` with a deliberately mismatched Vite identity and a non-default port (`3100`).

## PR Checklist:

Screenshot 2026-07-29 at 17 21 44


momustafa-causalens (approved):
i love that u created a whole PR to address my idiocy 🙏 
appreciate it

krzysztof-causalens:
> i love that u created a whole PR to address my idiocy 🙏 appreciate it

It's not just for you :P it happens to everyone and is a confusing failure mode all around so it's a good change regardless

PR #674, opened 2026-07-31
Fix: DO-12420 improve StreamVariable SSE reliability

krzysztof-causalens:
## Motivation and Context

A browser-facing `StreamVariable` can sit behind two separate SSE hops:

```text
API Server --(: keepalive comments + data)--> Dara --(data)--> browser
```

The API Server already keeps its upstream connection to Dara alive with SSE protocol comments, but Dara's intermediate SSE parser consumes those comments. The independent Dara-to-browser hop can therefore remain completely silent while the upstream stream is healthy. A proxy or browser-facing intermediary may then close that idle response cleanly. Before this change, clean EOF was treated as successful completion, so the client did not reconnect and the UI could remain stale.

This follows the StreamVariable disconnect cleanup work in #645. The two protections serve different purposes: browser-facing heartbeats prevent avoidable idle-path closures, while bounded reconnect remains the safety net when a closure actually occurs.

## Implementation Description

- Treat unexpected clean StreamVariable EOF as retriable, with bounded exponential backoff.
- Surface fatal server events immediately, stop after retry exhaustion, reset the retry count only after successfully applied application data, and never retry an intentional abort.
- Ignore comment-only/empty SSE messages as transport heartbeats without logging browser errors, emitting application events, or changing reducer state.
- Emit generic SSE `: keepalive` protocol comments from Dara while the browser-facing stream is quiet, including during dependency resolution. The default interval is 15 seconds and can be configured from 1–30 seconds with `DARA_STREAM_KEEPALIVE_INTERVAL_SECONDS`.
- Run dependency resolution, async-generator iteration, and generator cleanup in one persistent producer task. This preserves task-local context and gives the browser-facing consumer a narrow, deterministic cancellation seam.
- Explicitly own the nested `_generate_stream` iterator with `contextlib.aclosing`, so cancellation while the producer is blocked publishing a buffered chunk awaits generator teardown and releases upstream HTTP subscriptions before the parent returns.
- Explicitly close forwarding iterators at the route and stream-tracking wrapper seams. A StreamVariable-specific `StreamingResponse` owner also closes its body in a shielded `finally`, covering the real ASGI disconnect race where Starlette cancels while `send()` is blocked after pulling an application item.
- Shield `run_stream` teardown at the task-ownership seam. Its queue reader, producer, and disconnect waiter are cancelled and deterministically awaited without forwarding repeated parent cancellation into a producer already awaiting a user generator's `aclose()`. Deferred parent cancellation is re-raised only after the upstream async context has finished closing its transport.
- Make browser disconnect win over a simultaneous producer result or failure, then cancel and await all pending producer/consumer tasks so generators and upstream subscriptions are not leaked.
- Keep request ownership explicit: request extras may customize headers and fetch behavior, but cannot replace the StreamVariable POST body or abort signal.
- Document the two-hop behavior and configuration, and update the Dara Core changelog.

None.

## How Has This Been Tested?

- Dara Core JS suite: 32 files, 353 tests passed.
- Focused StreamVariable JS suite: 75 tests passed, covering clean EOF reconnect, fatal events, retry exhaustion/reset, intentional abort, heartbeat parsing and reconnect interaction, invalid events, and abort-signal ownership.
- Focused Python StreamVariable suite: 49 tests passed, covering bounded heartbeat timing, heartbeat during dependency resolution, normal delivery, generator completion/cancellation, context preservation, disconnect cleanup, and repeated parent cancellation while upstream cleanup crosses multiple checkpoints. The real HTTPX regression lets the first ready frame send normally, disconnects while `run_stream` is waiting for the next item, and proves an immediate replacement request succeeds through the same capacity-one pool.
- Complete Dara Core Python suite: 860 tests passed.
- Python lint and format checks passed.
- Full JS lint and repository-wide JS formatting checks passed.
- Dara Core JS production build completed successfully.

## PR Checklist:

N/A — protocol and client reliability changes only.

### DIFF
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -4,7 +4,7 @@ title: Changelog
 
 ## NEXT
 
-- Fixed StreamVariable clean connection closures not reconnecting and fatal stream errors silently leaving stale values visible, and added configurable SSE keepalive comments for idle browser connections.
+- Fixed StreamVariable clean connection closures not reconnecting and fatal stream errors silently leaving stale values visible, and added configurable SSE keepalive comments with robust parsing and cleanup across the full stream lifecycle.
 - Fixed FastAPI telemetry cleanup errors when application construction fails by attaching instrumentation only after
   the application has been built successfully.
 
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -9,6 +9,7 @@ title: Changelog
 
 ## 1.29.6
 
+- Fixed StreamVariable clean connection closures not reconnecting and fatal stream errors silently leaving stale values visible.
 - Added a development-only server handshake that reports mismatched projects and supports overriding Vite's port with `dara start --dev-port` and `dara dev --port`.
 - Fixed `dara dev` to use the application's configured static files directory instead of always assuming `dist`.
 - Added opt-in, vendor-neutral OpenTelemetry traces, logs, and metrics through Pydantic Logfire, covering HTTP,
--- a/packages/dara-core/changelog.md
+++ b/packages/dara-core/changelog.md
@@ -4,12 +4,12 @@ title: Changelog
 
 ## NEXT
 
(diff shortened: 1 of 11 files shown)


momustafa-causalens (approved):
LGTM!

PR #677, opened 2026-08-03
Fix: DO-12420 add backpressure-aware polling

krzysztof-causalens:
## Motivation and Context

This is the top PR in the native GitHub stack:

1. #674 — StreamVariable reconnect/transport reliability
2. #677 — this polling-backpressure and route-cancellation change

This draft intentionally targets `codex/fix-stream-variable-reconnect` and depends on #674.

DerivedVariable and Python-component polling previously used fixed-rate `setInterval` invalidation. A new selector execution could start while the prior request was still pending, and the DerivedVariable RxJS `switchMap` only unsubscribed from the old Promise—it could not cancel the underlying fetch. When response latency exceeded the polling interval, browser requests and backend computations piled up and could later arrive in bursts.

Rapid navigation exposed a related ownership bug in the route loader. If its NDJSON stream had already resolved the template or actions, aborting the stream tried to reject those handles again and raised `Deferred already resolved`.

## Implementation Description

- Add one `Poller` shared by DerivedVariable and Python-component polling, keyed by rendered identity and request extras.
- Keep one poll request in flight per key. Busy ticks request at most one trailing run, and ordinary polls use fixed delay from request completion.
- Pause timers while the document is hidden and run once when it becomes visible.
- Apply bounded ±10% client jitter, exponential error backoff capped at 60 seconds, success reset, and `Retry-After` support.
- Give each request generation an owned `AbortController`; pass its signal through fetches and task-result requests, abort on final unmount/navigation, and allow React StrictMode to reacquire ownership for one macrotask.
- Let real dependency changes supersede slow polling at once. Generation checks block stale writes, while an older selector evaluation adopts the newest completed result.
- Replace the uncancellable RxJS Promise switch with a small per-selector startup delay that keeps started request generations separate.
- Make route-handle settlement status-aware. Abort now rejects only pending handles, cancels the NDJSON reader explicitly, and leaves already-settled template, action, DerivedVariable, and Python-component data intact.
- Keep manual refresh behavior unchanged and leave unrelated fixed-rate `useInterval` consumers alone.

## Before/After Evidence and Impact

The old path emitted every second regardless of request lifetime. Holding a response open for more than ten intervals could start work on every tick.

The deterministic regression harness uses a one-second interval and holds the first poll response open:

- DerivedVariable: after ten more simulated seconds, total requests stayed at `2` (initial load plus the active poll) and peak poll concurrency stayed at `1`.
- Python component: the same case stayed at `2` total requests and peak poll concurrency `1`.
- Repeated ticks while busy lead to exactly one trailing run.
- Hidden tabs stop polling and run once after becoming visible.
- Active slow requests receive an aborted signal on final unmount.
- Rapid route changes can abort a partly read stream without double-settling resolved handles; the reader is canceled and the latest navigation wins.

Slow upstreams now apply backpressure instead of building browser, proxy, and backend queues. The UI keeps the last good value during a failed poll, refreshes promptly after visibility or input changes, and cannot be rolled back by a stale response. Normal navigation aborts no longer leak an uncaught `Deferred already resolved` error.

None.

## How Has This Been Tested?

- Focused deterministic suites: 4 files, 82 tests passed.
- Poller tests cover one active request and one trailing run, fixed delay, hidden-tab pause/resume, error backoff, success reset, `Retry-After`, bounded jitter, unmount/navigation abort, chained input changes, StrictMode ownership, shared-consumer reuse, and request-extras isolation.
- DerivedVariable and Python-component integration tests cover slow-request concurrency, suspended initial-request abort, input changes during a poll, stale-result protection, shared/nested consumers, and distinct request-extras.
- Route-loader tests cover abort after template/actions and some preload handles settle, abort before any chunks, pending-handle rejection, body cancellation, and latest-navigation wins.
- Complete Dara Core JS suite: 33 files, 383 tests passed.
- Dara Core type checking and touched-file type-aware lint passed.
- Full repository JS lint and repository-wide JS formatting checks passed.
- Dara Core production build completed successfully.
- `git diff --check` passed.

## PR Checklist:

N/A — request scheduling and lifecycle behavior only.

### DIFF
--- a/packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
+++ b/packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
@@ -18,7 +18,7 @@ import ProgressTracker from '@/components/progress-tracker';
 import { FallbackCtx, VariableCtx, useRequestExtras, useTaskContext } from '@/shared/context';
 import { ErrorDisplay, isSelectorError } from '@/shared/error-handling';
 import { useRefreshSelector, useVariable } from '@/shared/interactivity';
-import { usePolling } from '@/shared/interactivity/polling';
+import { beginPollOwner, keepPollOwner, releasePollOwner, usePolling } from '@/shared/interactivity/polling';
 import useServerComponent, {
     getServerComponentRequestKey,
     usePollServerComponent,
@@ -280,6 +280,14 @@ function DynamicComponent(props: DynamicComponentProps): React.ReactNode {
 
     const { hasRunningTasks, cleanupRunningTasks } = useTaskContext();
     const variables = useRef<Set<string>>(new Set());
+    const [pollingOwner] = useState(() => Symbol('dynamic-component'));
+    const variableContext = useMemo(() => ({ pollingOwner, variables }), [pollingOwner]);
+    beginPollOwner(pollingOwner);
+
+    useEffect(() => {
+        keepPollOwner(pollingOwner);
+    });
+    useEffect(() => () => releasePollOwner(pollingOwner), [pollingOwner]);
 
     /*
         When this component unmounts, then cancel any pending tasks for this component. This is required because recoil uses
@@ -322,7 +330,7 @@ function DynamicComponent(props: DynamicComponentProps): React.ReactNode {
             onReset={onResetErrorBoundary}
         >
             <FallbackCtx.Provider value={{ suspend }}>
-                <VariableCtx.Provider value={{ variables }}>
+                <VariableCtx.Provider value={variableContext}>
                     <Suspense fallback={fallback}>{component}</Suspense>
                 </VariableCtx.Provider>
             </FallbackCtx.Provider>
--- a/packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
+++ b/packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
@@ -15,11 +15,15 @@ import { ErrorBoundary } from 'react-error-boundary';
 import DefaultFallback from '@/components/fallback/default';
 import { hasMarkers } from '@/components/for/templating';
 import ProgressTracker from '@/components/progress-tracker';
-import { FallbackCtx, VariableCtx, useTaskContext } from '@/shared/context';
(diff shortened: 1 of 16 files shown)


packages/dara-core/js/shared/interactivity/polling.tsx
```diff
@@ -0,0 +1,544 @@
...
+
+/**
+ * Register fixed-delay polling for a request identity.
+ *
+ * `refresh` only invalidates the corresponding selector. Request lifecycle,
+ * backpressure, retries, and cancellation remain owned by the coordinator.
+ */
+export function usePolling(key: string, intervalSeconds: number | undefined, refresh: () => void): void {
+    const refreshRef = useRef(refresh);
+    refreshRef.current = refresh;
```
krzysztof-causalens: This should be done in layout effect, I think we have useLatestRef hook for this?

packages/dara-core/js/shared/dynamic-component/dynamic-component.tsx
```diff
@@ -276,6 +280,14 @@ function DynamicComponent(props: DynamicComponentProps): React.ReactNode {
 
     const { hasRunningTasks, cleanupRunningTasks } = useTaskContext();
     const variables = useRef<Set<string>>(new Set());
+    const [pollingOwner] = useState(() => Symbol('dynamic-component'));
+    const variableContext = useMemo(() => ({ pollingOwner, variables }), [pollingOwner]);
+    beginPollOwner(pollingOwner);
```
krzysztof-causalens: Should probably be in a layout effect, don't do side effects in component body. Then release owner can be in the same one's return maybe?

Or is this to work with suspense? If so then we need to be really careful as this technically breaks react's rules, can we double check react's guidance if this is the most idiomatic solution.

packages/dara-core/js/shared/interactivity/polling.tsx
krzysztof-causalens, line 1: needs more inline comments in some intricate code blocks inside to explain the logic and choices

momustafa-causalens (approved):
LGTM!

PR #685, opened 2026-09-03
fix: DO-12796 - fix logout auth redirect race

momustafa-causalens:
## Motivation and Context

Stale API requests can return a 401 response after logout revokes the session. `handleAuthErrors()` then redirects from `/logout` to `/login` with `/logout` as the referrer. A successful login returns the user to `/logout` and starts the logout flow again.

This change lets the logout flow control navigation from `/logout`.

Repro:
App enters /logout
  → session is revoked
  → old derived-variable requests return 401
  → handleAuthErrors() reads window.location.pathname
  → current pathname is /logout
  → browser navigates to /login?referrer=%2Flogout
  → successful login returns to /logout
  → logout runs again

## Implementation Description

- Keep the transition active until the final navigation completes.
- Allow the navigation callback to return `void | Promise<void>`.
- Await the navigation callback before releasing logout ownership.
- Await React Router navigation in the OIDC logout flow.

The shared helper intentionally replaces separate `beginLogout()` and `finishLogout()` calls. This keeps the required ordering in one place and prevents callers from ending the transition too early or forgetting cleanup. Custom logout components can use the same helper without coupling authentication handling to their route paths.

None.

## How Has This Been Tested?

- Verify that the transition remains active after revocation completes while navigation is pending.
- Verify that a stale authentication failure remains suppressed during this window.
- Verify that the transition ends only after the navigation promise settles.
- Ran `./node_modules/.bin/vitest --run tests/js/auth.spec.tsx` from `packages/dara-core`. All 26 tests passed.
- Ran `./node_modules/.bin/tsc --noEmit` from `packages/dara-core`.
- Ran ESLint on `js/auth/auth.tsx` and `tests/js/auth.spec.tsx`.
- Ran Prettier checks on `js/auth/auth.tsx` and `tests/js/auth.spec.tsx`.
- Ran `git diff --check`.

## PR Checklist:

Not applicable.


krzysztof-causalens (changes requested):
Good catch, thanks for investigating! I think I'd prefer a different approach to fixing it though:
1. We're tying the core auth logic to a hardcoded `/logout` URL so indirectly coupling to it. 
2. The above also doesn't work if for some reason we land on a different casing, e.g. `/logOUT` which the router normalizes but we compare literally (tho it's just an edge case)
3. Dara allows plugins/users to add other custom auth routes as part of auth configurations so again the hardcoded list of routes doesn't work as it isn't extentible. 

The alternative I'm thinking of is simply doing something like

```tsx
// session-state.tsx
type AuthTransition = 'idle' | 'logging-out';

let authTransition: AuthTransition = 'idle';

export function beginLogout(): void {
    authTransition = 'logging-out';
    notifySessionLoggedOut();
}

export function finishLogout(): void {
    authTransition = 'idle';
}

export function isLoggingOut(): boolean {
    return authTransition === 'logging-out';
}
```

then `handleAuthErrors` is independent of the routes. Each logout component would enter this state before revoking the session and leave it only after initiating its final navigation. 
Just need to be careful to mark the state transitions in the right places, can maybe wrap transition and logout operations in some shared helper so we don't even have to expose the state above. What do you think?

krzysztof-causalens (changes requested):
`runLogout()` stops owning auth navigation before React Router navigation has completed. `onLoggedOut` is typed as synchronous and called without `await`, while the OIDC callback explicitly discards the `void | Promise<void>` returned by `navigate('/login')`. With `createBrowserRouter`, the location is committed asynchronously, so the `finally` block can decrement `activeLogoutTransitions` while the browser is still on the logout route. A stale 401 arriving in that window sees `isLoggingOut() === false` and can start the competing login redirect again.

Could we allow the navigation callback to return `void | Promise<void>`, await it inside `runLogout()`, and await `navigate('/login')` in the OIDC callback? The regression test should use a deferred navigation promise, finish revocation, then verify that a stale auth failure remains suppressed until navigation settles. The current tests only exercise a synchronous callback or a 401 received while revocation is still pending, so they do not cover this window.

krzysztof-causalens (approved):

PR #692, opened 2026-09-08
Breaking: DO-0 support shared workspaces and library exports

krzysztof-causalens:
## Motivation and Context

Fourth slice of #623, stacked on #691. Two Dara apps sharing a pnpm workspace must preserve each other's dependency requirements, and a library should work from source locally and from compiled exports after publication.

## Implementation Description

Preparation uses pnpm's declared workspace membership and a typed snapshot of package declarations and recorded Python requirements. Catalog conflicts name the affected apps before any edit; Python package upgrades refresh requirements without importing sibling apps. Dependency drift is checked with pnpm's frozen verification in a disposable metadata mirror, and filtered installation covers the app's dependency graph.

The plugin validates configuration, builds workspace dependencies, then resolves registered implementations and checks TypeScript, so `dara lock` and production builds work from an unbuilt library. Sibling sources enter type checking through a private declaration file that preserves the app's TypeScript project. Workspace libraries keep source exports in the checkout and use `publishConfig.exports` for compiled publication. `dara build --no-deps-build` skips dependency scripts.

`rehype-raw` moves to v7 to match the Markdown renderer's unified/vfile types.

## How Has This Been Tested?

- Full repository build, Python and JavaScript suites, lint and format checks.
- Plugin tests cover real pnpm dependency-build ordering, native TypeScript checks, export resolution and output validation. Python regressions cover membership, catalog conflicts and upgrades, dependency drift and discovery cancellation.
- A fresh two-app workspace prepares without compiled library output, both apps build independently, and the packed library works in a separate Dara consumer and a plain JavaScript consumer.

## PR Checklist:


packages/vite-plugin/src/workspace.ts
```diff
@@ -0,0 +1,184 @@
...
+      "workspace.target",
+      `${reference} must target a file or package inside ${workspace}`,
+      "edit the local dependency reference",
+    );
+  }
+  return fs.realpathSync(target);
+}
+
+/** Inspect reachable repository packages and validate local references against their real targets. */
+export function workspaceGraph(root: string, workspace: string) {
```
krzysztof-causalens, line 60: Just wondering, could pnpm be used as a library here?

PR #694, opened 2026-09-08
Breaking: DO-0 complete Vite release integration

krzysztof-causalens:
## Motivation and Context

Final slice of #623, stacked on #693. CI and release packaging exercise the app-root Vite workflow end to end, and Dara no longer ships or invokes the legacy UMD pipeline.

## Implementation Description

Core and components publish compiled ESM through npm and build Python wheels with `uv build` into the repository-root `dist/`. Native pnpm publishing applies `publishConfig.exports` and publishes in dependency order. Legacy generated assets, scaffolding, templates, the old dev server, HTML injection settings and the constant `build_mode` bootstrap field are removed; removed settings and commands report their replacement and point to the migration skill.

`tooling/scripts/check_frontend_release.py` packs the npm packages and wheels, installs them in a fresh environment and exercises two workspace apps, generator output, a migrated legacy app, a packed library consumer and a plain JavaScript consumer, including browser rendering and HMR. CI runs it next to Cypress. Release versioning refreshes the full lockfile and then reconciles app catalogs. Development shutdown drains in-flight configuration refreshes before closing watchers.

Deployment copies the complete output, including `.dara-build.json`, into `dist/`; artifact-only runtimes need neither frontend sources nor JavaScript tools. The release action gets its own Dara 2 major; this PR supplies the build/start interface.

None. `fastapi-vite-dara` is removed.

## How Has This Been Tested?

- All package builds, JavaScript test targets and Python suites; lint, format and security checks; `mise run lock-frontends`.
- The packed release matrix passed, including migration of a legacy app, frozen dependency files, credential exclusion and artifact-only startup with an empty `PATH`. Cypress passed against the production build.
- A deterministic regression reproduces the leaked-watcher shutdown hang and passes with the fix. A dry-run registry rehearsal verified the publication recipe with zero writes.

## Development frontend page

`dara dev` serves a placeholder at `/` whenever the frontend is not ready. It was unstyled markup with no charset, no viewport and an always-present empty `<pre>`. It now matches the authentication screens (same radial gradients and card), supports light and dark via `prefers-color-scheme`, and honours `prefers-reduced-motion`.

It cannot reference anything under `/static`, since the frontend is exactly what is unavailable, so all CSS is inline and it uses a system font stack.

Four distinct states, where the previous page collapsed the first three into one:

| Situation | Heading | Suggested command |
| --- | --- | --- |
| `dara dev` still preparing | Waiting for the frontend | — |
| `dara dev --backend-only` | Backend-only mode | `dara dev --frontend-only` |
| Served without `dara dev` (IDE run config, direct `uvicorn`, orphaned backend) | No frontend is running | `dara dev` or `dara start` |
| Runner reported a diagnostic | Frontend blocked | the diagnostic's own fix |

Only the genuine waiting state animates; a blocked or idle project is not progressing, so its indicator sits still and turns red for blocked.

Waiting for the frontend

Waiting for the frontend in dark mode

Backend-only mode suggesting dara dev --frontend-only

No frontend is running

Frontend blocked showing a diagnostic and its fix

## PR Checklist:


packages/dara-components/pyproject.toml
```diff
@@ -42,8 +42,8 @@ dara-core = { workspace = true }
 
 [tool.hatch.build.targets.wheel]
 packages = ["dara"]
-# Generated JS bundles are gitignored; ship them anyway (poetry `include` equivalent).
-artifacts = ["dara/components/_assets/auto_js/*"]
+# Wheels ship only the committed common assets; legacy auto_js bundles stay out.
+exclude = ["dara/components/_assets/auto_js"]
```
krzysztof-causalens: Why is auto_js folder even a thing anymore? Shouldn't this be removed?

packages/dara-core/docs/advanced/custom-js.mdx
```diff
@@ -2,656 +2,168 @@
...
+  return <meter min={0} max={100} value={value}>{value}</meter>;
 }
 ```
 
-In this example, the content can be passed as a Variable or not and the hook will process what it receives to ensure
-that the `content` variable is set correctly. Further to this, the second element of the returned array is a setter
-function that allows for the variable to be updated, in the example we added an input that can be used to control the
-content. This setter will propagate any changes to all other uses of `useVariable` that are subscribed to the same
-`Variable` element on the python side.
+Import and use `Gauge` in your Python pages as any built-in component. Existing import discovery registers it. For dynamically imported classes, use `config.add_component(Gauge)`. A barrel export and a `local=True` flag are unnecessary. A local path must start with `./js/` and cannot escape that directory. The default export can have any JavaScript name; Python's component identity remains unchanged.
```
krzysztof-causalens: Don't even mention legacy flags like `local=True`, just treat this as a new clean slate

packages/dara-core/docs/advanced/custom-js.mdx
```diff
@@ -2,656 +2,168 @@
...
-    print(f'Received custom message: {message} from channel {str}')
-    # response is sent back to the client, must be JSON-serializable
-    return {'response': 'Hello from Python!'}
-
-config.add_ws_handler(kind='my_custom_kind', handler=handle_custom_message)
+```ts
+plugins: [dara({
+  inputs: ['build/banner.txt'],
+  directories: ['build/templates'],
+  environment: ['VITE_PRODUCT_VARIANT'],
```
krzysztof-causalens, line 130: Should explain exactly what those fields do

packages/dara-core/docs/advanced/custom-js.mdx
```diff
@@ -2,656 +2,168 @@
...
-
-# Get the channel associated with the user; this lookup will only succeed if the user is currently connected
-session_id = sessions_registry.get(user_id)
-channel_id = websocket_registry.get(session_id)
-
-# Send a message to the channel
-ws_manager = utils_registry.get('WebsocketManager')
-ws_manager.send_message(channel_id, {'kind': 'my_kind', 'data': 'some_payload'}, custom=True)
-```
+Script ordering and generated library tags are replaced by module imports. Genuine vendored files such as plotting libraries remain static assets and are loaded by their components. Development watches static roots; conflicting destinations produce a diagnostic and resume serving after the collision is fixed.
```
krzysztof-causalens: Again keep it simple and explainable for the user, not dense prose

packages/dara-core/changelog.md
krzysztof-causalens, line 5: The whole section about build changes should probably be consolidated a bit, some less relevant ones like DownloadContentImpl can be removed, same with `__proto__` as those are internal. Could collapse into ~4 important higher level items. Details can be in the migration skill.

